PR #22, opened 2022-08-24
Add flake8, a linter.

n-rook:
This will help us set a consistent style. I was annoying today and came in
to post a bunch of comments on a PR. But I live in the States... This way
it is less likely I will have annoying comments.

Really, I do not even care that much about most of these requirements.
It is just good to have a consistent style one way or another, and
some of the things it catches (like "unused module") are clearly good
to have. If anyone hates one of these rules we can probably get rid
of that one.


n-rook:
I also deleted a bunch of files that were unused. The people who complained about them were right :p

PR #26, opened 2022-08-25
th07 support

raviddog:


raviddog:
i havent put in proper shot ids or game name because we're changing game_ids anyway and i wanted to focus on making it work

n-rook (approved):

PR #34, opened 2022-08-29
i renamed the entire module from scores to replays im sorry

raviddog:
it will need a database reset im sorry
gotta delete and create and then rerun migrations and constant tables again im sorry


project/thscoreboard/replays/models.py
```diff
@@ -89,46 +89,69 @@ class Meta:
...
+    points = models.BigIntegerField()
+    """The final score of the run. If the run counterstops, this value should contain the post-counterstop score"""
+
+    #   replay metadata
+
     shot = models.ForeignKey('Shot', on_delete=models.PROTECT)
     """The shot type the player used."""
 
-    difficulty = models.IntegerField()
+    rep_date = models.DateTimeField()
```
n-rook, line 110: You put way too many changes in this PR. Why are we renaming the scores app and the table, but then adding a bunch of random fields at the same time? It is true that this PR requires we blow up the database, but it is not like we need to wait until the database gets blown up to add these fields and tables. We can always add them later instead.

Please split this PR up into only renaming the app and other stuff. Don't add extra columns or tables at the same time.

Actually there is one change that would be useful to do at the same time: Changing the "ReplayFile" table to only have the replay data itself, so that we never have to fetch it in order to do stuff like display the replay table or the replay details page. So if you want to do that too it is fine.

In general, I also don't think it's a good idea to add extra columns and tables we aren't using yet. The reason is that if we never end up using them, they will just waste code and space in the database. And if we decide they should have slightly different names or types, it will be annoying to change them later. Instead I think it is better to add the extra columns and tables when we actually implement the feature around them. For example you could add fields like `slowdown` or `clear` at the same time as adding code to the replay parser to actually return and save that code.

project/thscoreboard/replays/models.py
```diff
@@ -89,46 +89,69 @@ class Meta:
...
+    """ The final score recorded in the replay.
+    
+        This will usually be the same as the score on the Score row, but in some
+        cases it will be different. For example, this will be the max score for
+        counterstop replays.
+    """
+
+    rep_name = models.TextField()
+    
+    rep_difficulty = models.IntegerField()
```
n-rook, line 126: I don't really like this "rep_" naming convention. I don't think we really get anything out of having a naming convention to verify which fields came from the replay and which fields didn't. Plus, a lot of these fields are relevant to videos too.

project/thscoreboard/replays/models.py
```diff
@@ -141,13 +164,85 @@ def GetNiceFilename(self, ascii_only=False):
...
+    th17_token3 = models.IntegerField(blank=True, null=True)
+
+    th17_token4 = models.IntegerField(blank=True, null=True)
+
+    th17_token5 = models.IntegerField(blank=True, null=True)
+
+
 class ReplayFile(models.Model):
-    """Represents a replay file for a given score."""
+    """Represents a replay file or video for a given submission."""
```
n-rook, line 240: We wanted to make this table smaller, right? That way, we can less often have to fetch it, which will make the site faster and make it put less load on the database. As such, I don't think it is a good idea to move additional data to it. Instead I would rather see us move even "points" (presumably as "replay_file_points" or "true_points" or something) and "is_good" to the main table (or to another new table).

n-rook (commented):
I will never forgive you!!! 😭😭😭

I still think this is a huge waste of time but everyone else wants it so OK. But this PR has way more in it than just renaming the folder so please make it smaller. (I discussed this more in another comment)

I am sorry that it is inconvenient to go back and forth on reviews. The time zone is annoying since I really can't review PRs during my real job, but I am sure it is annoying.

raviddog:
i have no idea how to create a PR that doesnt automatically dump all my latest commits into it
am i just supposed to do the branch method and wait for each PR to be merged

project/thscoreboard/replays/models.py
raviddog: whoops
i was thinking it'd be a nice idea to keep the sources of the replay together, but you're right it doesn't make sense

project/thscoreboard/replays/models.py
raviddog: I actually did only rename the app and table, then submitted the PR, then I did extra work and that got automatically added to the PR and idk how to not do that

raviddog: the extra columns are there because I know we will be using them because we've already implemented all those in the replay parsers and all that's left is to hook it up

PR #35, opened 2022-08-30
rename scores module to replays

raviddog:
attempt no 2
today i learned why people constantly branch and merge
this PR should only contain stuff to make the repo work after resetting the database


n-rook (approved):

n-rook:
oops, I merged in the wrong order. you will have to fetch the new updates to fix this, I don't think I can fix it on my end

raviddog:
ok i thought this would be harder but it was pretty simple

PR #38, opened 2022-08-31
Add translation.

n-rook:
The translation is basically empty at this point, but the hooks are in now. It should at least be possible to mark strings for translation going forward.

(There is also not any way to actually switch to Japanese yet...)


n-rook:
I will hold off on merging this until #35 is in.

PR #39, opened 2022-09-02
Table stage splits

raviddog:


project/thscoreboard/replays/models.py
```diff
@@ -141,6 +175,78 @@ def GetNiceFilename(self, ascii_only=False):
...
+
+    bombs = models.IntegerField(blank=True, null=True)
+
+    bomb_pieces = models.IntegerField(blank=True, null=True)
+
+    graze = models.IntegerField(blank=True, null=True)
+
+    point_items = models.IntegerField(blank=True, null=True)
+
+    """Game specific entries"""
```
n-rook, line 203: Nitpick: There is a difference between "docstrings" and "comments." Docstrings are a Python syntactic concept where putting a string in specific places attaches it as documentation to a nearby entity. See https://peps.python.org/pep-0257/#what-is-a-docstring.

This line is not really a docstring; it just categorizes a number of fields and serves as a header. So it should really just be a comment. As such, you should say it as

```
# Game-specific entries.
```

project/thscoreboard/replays/models.py
```diff
@@ -141,6 +175,78 @@ def GetNiceFilename(self, ascii_only=False):
         )
 
 
+class ReplayStages(models.Model):
```
n-rook, line 178: I mentioned this on the last PR, but I don't think it's a good idea to add a bunch of unused database fields. Doing so has a few problems:

- If we do not end up setting them at the end, they just sit around being distracting.
- Similarly, they can be confusing to a developer who expects them to be set.
- It is not clear exactly what their meanings are.
- Many of them have to be nullable even if they will be set.
- We might be surprised by the values of certain fields, or their types.

As such, I really strongly think we will be better off splitting this into multiple PRs. I would probably recommend:

* Make the changes to the base replay table that don't require anything else, like renaming "points" to "score".
* We can make the replay parser return more data first. In some sense this is "useless code" too, but the good part about it is that the code still works, and isn't just empty values.
* (It would also be OK to parse the information out of the replay in the PR before sending it in. But this might end up being too big of a PR to be convenient, since you would have to get the data from multiple replay formats.)
* Then, once we have the information, we can start putting it into the tables.

For doing this, it's fine to actually send multiple PRs in parallel. For example there is no reason renaming "points" to "score" or deleting some of the columns of ReplayFiles needs to depend on adding stage splits, or vice versa. So it should not add too much time. If anything it will make it take less time to get things in, because there will be less back-and-forth on smaller PRs.

project/thscoreboard/replays/views.py
```diff
@@ -355,23 +355,24 @@ def GetReplayOr404(user, replay_id):
...
         shot=shot,
         difficulty=difficulty,
-        points=points,
+        score=score,
+        rep_score=score,
         category=category,
         comment=comment,
-        video_link=video_link,
+        date=datetime.datetime.now(tz=None),
+        name="temptest",
```
n-rook, line 370: This is the sort of thing which is why I want us to parse the information out of the replay file first. If we did that, it would be very easy to just pass the right information in here, and we wouldn't have to pass fake default info.

project/thscoreboard/replays/views.py
```diff
@@ -380,16 +381,18 @@ def PublishNewReplay(user, difficulty: int, shot: models.Shot, points: int, cate
...
         user=user,
         shot=shot,
         difficulty=difficulty,
-        points=points,
+        score=score,
+        rep_score=score,
         category=category,
         comment=comment,
+        date=datetime.datetime.now(tz=None),
+        name="temptest",
```
n-rook, line 394: We should probably not record the name "temptest" for replays without replay files, or a fake date record. I think it'd be better to just record no data for these columns.

project/thscoreboard/replays/models.py
n-rook, line 178: There has to be more documentation on this table; I don't know what half these fields are! Are things generally the state of the game at the start of the stage, or the end? What's the difference between "lives" and "extends"? Which info is defined for which games?

(This is a big part of why I want us to put real data in these columns when we add them; it'll be easier to know the answers to these questions that way.)

n-rook (changes requested):
Thanks!

raviddog:
thanks for the detailed responses, this is both my first python project and my first real time using pull requests extensively so i hope my beginner mistakes aren't too annoying

n-rook:
No, it's fine. Certainly these contributions are very helpful. Really I worry the other way around, that I am making it too difficult for other team members to contribute, but I do think this is the best way to get a high-quality and maintainable codebase.

raviddog:
Its certainly a bit rough starting out but teaching us to make proper contributions will ultimately speed things up in the future

PR #46, opened 2022-09-06
move is_good and points out of replayfile table

raviddog:
i hope django's smart enough to not actually try to pull the data from the table anymore with this change


project/thscoreboard/replays/models.py
```diff
@@ -109,6 +109,21 @@ def GetDifficultyUrlCode(self):
 
     points = models.BigIntegerField()
     """The score of the replay."""
+    
+    is_good = models.BooleanField(blank=True, null=True)
+    """Whether a replay file can be used unmodified to watch the replay.
+    
+    Even a desynced replay file can be useful to have. For example, maybe the
+    Touhou community will later discover how to fix a certain type of desync.
```
n-rook, line 117: It's worth noting that this field will be unset if the replay doesn't have a replay file.

n-rook (approved):
Make sure to check creating non-file replays works before submitting this change.

PR #47, opened 2022-09-06
add replay timestamp

raviddog:
pull requests that have database modifications probably need to be merged in the right order otherwise im scared of what happens to the migrations

also i tried to split this into two PRs but why does it still include the commit from the other one
idk what to do tbh


project/thscoreboard/replays/views/create_replay.py
```diff
@@ -212,7 +214,9 @@ def PublishReplayWithoutFile(user, difficulty: int, shot: models.Shot, points: i
         points=points,
         category=category,
         comment=comment,
-        video_link=video_link
+        video_link=video_link,
+        timestamp=datetime.now(),
```
n-rook, line 218: IMO we should leave this field as null for non-file replays: After all, we don't actually know when the user played the original run. We have the "created" field to record when the user uploaded the replay.

n-rook (changes requested):
To deal with the previous change, I would recommend getting rid of that commit on this branch. After all, this change should not really depend on that one. The way I would do this is with `git rebase -i`, the "interactive rebase" command.

Basically, `git rebase -i` lets you change history. This is a really bad idea if you are working on a "real" branch that other users are also fetching, but it is fine if you are working on a branch like "rpy_timestamp" that only exists for you to make a pull request. This looks like decent documentation on it: https://docs.gitlab.com/ee/topics/git/git_rebase.html#interactive-rebase

You can use this command to excise the previous commit from this branch. Obviously, this would not be a good idea if you didn't already have the previous commit on another branch, since you would lose it. But since you do, it's fine.

The only complication when it comes to one change depending on another is the migration. It actually is fine to merge migrations in a weird order:

https://docs.djangoproject.com/en/4.1/topics/migrations/#version-control

The numbers in the filenames are just for us, they don't actually mean anything to Django. But the "dependencies" field does list the other PR's migration explicitly, which would cause problems. Read here for more information:

https://docs.djangoproject.com/en/4.1/topics/migrations/#dependencies

I think it would work to just replace its name in the migration file with the initial migration (that is, `0001_initial`). I think you could also fix this by first removing the previous commit, then reversing the migration (see https://docs.djangoproject.com/en/4.1/topics/migrations/#reversing-migrations), then deleting the migration file, then regenerating it by running `makemigrations` again.

Of course... in reality you don't need to do anything because I just approved the other PR... But I think this stuff is good to know in general.

n-rook (approved):

n-rook:
Something weird is going on with the diff. If you look at "files changed" it looks horrible. But then I went and actually double-checked the commits and it looks completely fine...

PR #48, opened 2022-09-07
"Make the forms look a bit nicer" now with tables

32th-System:
This reverts commit d6a3ad317de79aaaf052a2d7c6724ff54307594e.

Turns out that when the amount of things on the left side is not equal to the amount of things on the right side, nothing on either side aligns with eachother. Ideally, if something extra appears on the left, it would push everything below it down and make sure it still aligns. There is a tool that does just that, *t a b l e s*. I realized this right as my previous pull request got merged but only now did I fix this

### DIFF
--- a/project/thscoreboard/replays/templates/replays/base_publish_replay.html
+++ b/project/thscoreboard/replays/templates/replays/base_publish_replay.html
@@ -2,45 +2,79 @@
 
 {% block content %}
 {% block pre_replay_form %}{% endblock %}
-<form method="post" class="form-flex">
+<form method="post">
     {% csrf_token %}
     {{ form.non_field_errors }}
 
-    <div class="form-body" style="max-width: 30em;">
-    <div class="form-block">
-        {{ form.difficulty.errors }}
-        <label for="{{form.difficulty.id_for_label}}">Difficulty</label>
-        {{ form.shot.errors }}
-        <label for="{{form.shot.id_for_label}}">Shot</label>
-        {{ form.points.errors }}
-        <label for="{{form.points.id_for_label}}">Score</label>
-        {{ form.category.errors }}
-        <label for="{{form.category.id_for_label}}">Category</label>
-        {{ form.video_link.errors }}
-        <label for="{{form.video_link.id_for_label}}">Video link</label>
-
-        {% if has_replay_file %}
-        {{ form.is_good.errors }}
-        <label for="{{form.is_good.id_for_label}}">No desyncs</label>
-        {% endif %}
-        
-        {{ form.comment.errors }}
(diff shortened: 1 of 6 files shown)


project/thscoreboard/replays/templates/replays/base_publish_replay.html
```diff
@@ -2,45 +2,81 @@
...
-    </div>
-    
-    <div class="form-block">
-        {{ form.difficulty }}
-        {{ form.shot }}
-        {{ form.points }} 
-        {{ form.category }} 
-        {{ form.video_link }}
+    <table>
+    <tbody style="vertical-align: top;">
```
n-rook, line 10: We should avoid inline style tags, I think. From what I can tell, they are not considered to be a good idea in the web development community. And I can kind of see why; the nice thing about a class is it makes it really easy to share code from one place to another. In this case you are putting the same "vertical-align: top;" instruction on every table body; I think it would be better to do this in CSS somehow so it's easy to see all the tables are styled the same way, and so that it's easy to change table styling further if we do so later.

project/thscoreboard/replays/templates/replays/base_publish_replay.html
```diff
@@ -2,45 +2,81 @@
...
+        <th style="white-space: nowrap; width: 50%;"></th>
```
n-rook, line 11: Does `white-space: nowrap;` actually do something here?

n-rook (changes requested):
Thanks!

project/thscoreboard/replays/templates/replays/base_publish_replay.html
32th-System: Yes. The whole point of that line is is forcing a particular minimum width for the first column to add space between it's elements and the second column's elements

project/thscoreboard/replays/templates/replays/base_publish_replay.html
32th-System: I used an inline style because I thought that it's OK to do if there's only one line of styling. Then there's `<th style="white-space: nowrap; width: X;"></th>`, where `X` is sometimes 50%, and sometimes 100%. This is something I'd much rather inline if I'm specifically using different values in different places

n-rook: I think it's good to use CSS even for one-line changes because they make it easy to add additional changes to the same class of entity later. I am kind of ambivalent about the `width` tag. I think if I were doing it I would still use CSS but I can't really come up with a good argument for or against it.

n-rook (approved):

PR #50, opened 2022-09-14
Add server files

32th-System:


n-rook (approved):
Looks mostly good but let's put these somewhere else, not just in ref/. I would probably put the nginx conf actually in a folder named something like nginx/. maybe a top level folder called "server", which would then contain silentselene.service and also an nginx folder

PR #51, opened 2022-09-15
Specify a "Real Name" in the email headers

32th-System:
Turns out it's pretty easy to do. The way the django documentation documents it is wrong. It's not `"Fred" <fred@example.com>`, it's `"Fred <fred@example.com>"`, but really, the way I figured it out is by looking at the headers from some email in my inbox and comparing with the Silent Selene emails


n-rook (approved):
Nice.

For stuff like this, I always go straight for the RFC!

PR #54, opened 2022-09-18
th06 stage splits

raviddog:


project/thscoreboard/replays/kaitai_parsers/th06.py
```diff
@@ -74,13 +74,16 @@ def stages(self):
...
         self._m_stages = []
         for i in range(7):
             _on = self.header.stage_offsets[i]
             if _on == 0:
                 self._m_stages.append(Th06.Dummy(self._io, self, self._root))
             else:
+                #   the th06 decryptor only returns the decrypted data instead of the full file
+                #   the game stores these stage offsets from the start of the file
+                #   thus we must adjust them to account for this difference when reading the stage data
+                self._io.seek(self.header.stage_offsets[i]-15)
```
n-rook, line 86: I don't think we should have to implement the -15 ourselves. In theory these source files should require no customization after being generated from the Kaitai Struct files, right? So I think it would be better to apply this -15 there. I don't know KS well and don't have a compilation chain set up yet, but I think something as simple as `pos: header.stage_offsets[_index] - 15` would work, right?

project/thscoreboard/replays/kaitai_parsers/th06.py
```diff
@@ -74,13 +74,16 @@ def stages(self):
             return self._m_stages
 
         _pos = self._io.pos()
-        self._io.seek(self.header.stage_offsets[i])
```
n-rook, line 77: We should file an issue about the fact that we need to do these manual interventions in the Kaitai Struct generated code. It would be best if we didn't have to modify these source files at all. Could you file a bug for this?

The issue on our side would have two possible resolutions:
- Work around this issue, or if it's caused by a bug in our struct definitions, fix the bug. (It doesn't seem like there's anything wrong with our definitions, so this seems kind of unlikely)
- Figure out a minimal reproduction of the bug, file a bug against Kaitai Struct, and maybe even send them a pull request to fix the bug

But really, fixing it would not be a high priority or anything. I just want a bug so we don't forget this is a thing.

project/thscoreboard/replays/models.py
```diff
@@ -163,6 +163,39 @@ def GetNiceFilename(self, ascii_only=False):
         )
 
 
+class ReplayStage(models.Model):
```
n-rook, line 166: I think for the model, we should document exactly which games the fields belong to, at least for the shared fields (that is... it goes without saying that `th06_rank` is only set for th06). We should also note if the field is always set for a game or not.

For example we might say:

graze = ...
"""The number of grazes the player has made at the beginning of the stage.

Set for:
- th06 (required)
- ...
```

Of course, for now you would just say "th06" for everything and not cover other games, since it is the only one implemented. But it will make it easier to keep track of later.

File: project/thscoreboard/replays/models.py [hunk 4]
```diff
@@ -163,6 +163,39 @@ def GetNiceFilename(self, ascii_only=False):
...
+    """Represents a stage split for a given replay"""
```
n-rook, line 167: Do the fields in this model refer to the state at the beginning of the stage, or at the end? I think I know the answer but we should still note it down somewhere.

project/thscoreboard/replays/models.py
n-rook, line 166: We should document every field. I know even I haven't been completely diligent about this, but I do think it's the way to go. The database is always the most important part of the codebase for people to know exactly what each field or row means. After all, if there is a bug in code it is usually easy to just fix and deploy it, but if bad data gets into the database it can be much more annoying to track down and fix.

IMO the "hierarchy of docs" is as follows:
1. Anything used as an interface between the server and something outside the server. This includes APIs and the database. These are the most important to document, because they tend to be used in many places. Plus, it is difficult to tell what the definitions are just by looking at code; you generally have to look in lots of different places to figure it out. (In theory HTML views would be included here... but in reality it is basically impossible to document those well since they're big documents, they're not something like a JSON response that is easy to describe.)
2. Any public function, method, etc. in a module. These are still important to document, because it's good for coders to be able to use them without having to actually read their implementation. But it's less important than (1) because you generally only need to read the function definition and maybe its tests to figure out mostly how it works.
3. Anything that isn't public in a module. I generally don't bother documenting these (or use a one-line docstring) unless they are unusually complicated.

project/thscoreboard/replays/models.py
```diff
@@ -163,6 +163,39 @@ def GetNiceFilename(self, ascii_only=False):
...
+    class Meta:
+        constraints = [models.UniqueConstraint(fields=['replay', 'stage'], name='unique_stage_per_game')]
+
+    replay = models.ForeignKey('Replay', on_delete=models.CASCADE)
+    """The replay this split corresponds to"""
+
+    stage = models.IntegerField()
+    """The stage this split corresponds to in the replay"""
+
+    score = models.BigIntegerField(blank=True, null=True)
```
n-rook, line 178: `score` shouldn't be nullable, right?

project/thscoreboard/replays/models.py
```diff
@@ -163,6 +163,39 @@ def GetNiceFilename(self, ascii_only=False):
...
+    """Represents a stage split for a given replay"""
+
+    class Meta:
+        constraints = [models.UniqueConstraint(fields=['replay', 'stage'], name='unique_stage_per_game')]
+
+    replay = models.ForeignKey('Replay', on_delete=models.CASCADE)
+    """The replay this split corresponds to"""
+
+    stage = models.IntegerField()
+    """The stage this split corresponds to in the replay"""
```
n-rook, line 176: Does this start at 0 or 1? What is the stage number for extra? (I think I know the answer to these, but for database stuff it's best to be really explicit.)

project/thscoreboard/replays/replay_parsing.py
```diff
@@ -50,15 +66,33 @@ def _Parse06(rep_raw):
...
-    return ReplayInfo(
+    r = ReplayInfo(
         game_ids.GameIDs.TH06,
         shots[rep_raw[6]],
         rep_raw[7],
         replay.header.score,
         datetime.strptime(replay.header.date, "%m/%d/%y")
     )
 
+    r.stages = rep_stages
```
n-rook, line 92: Nitpick: I think it's slightly better style to pass it into the ReplayInfo constructor directly.

project/thscoreboard/replays/replay_parsing.py
```diff
@@ -50,15 +66,33 @@ def _Parse06(rep_raw):
     replay = th06.Th06.from_bytes(cryptdata)
     
     shots = ["ReimuA", "ReimuB", "MarisaA", "MarisaB"]
+
+    rep_stages = []
+    i = 0
+    for score in replay.stages:
```
n-rook, line 72: We should add tests for all these changes. They should be pretty easy to write.

n-rook (changes requested):
Thanks, it'll be great to have this in!

project/thscoreboard/replays/replay_parsing.py
raviddog: i already have a separate branch set up that i plan to create a PR once this gets merged

project/thscoreboard/replays/kaitai_parsers/th06.py
raviddog: i agree, maybe create an issue so we don't forget, but i just wanted this stupid thing working so i could actually make progress

project/thscoreboard/replays/kaitai_parsers/th06.py
raviddog: see comment about -15

project/thscoreboard/replays/models.py
raviddog: theres so much variation here that im kinda scared to do it as code comments because itll bloat the file and make the lines really far from each other. i want to be able to see all the fields in this table at a quick glance, so yeah idk.

project/thscoreboard/replays/models.py
raviddog: the beginning, except for th07 which stores the end because zun is inconsistent. we could "fix" this because we know what values you start with but idk, raise an issue so we can discuess fuether i guess

project/thscoreboard/replays/models.py
raviddog: i dont trust zun
but yeah youre right, and we can always change this

project/thscoreboard/replays/replay_parsing.py
raviddog: i would have to manually declare the __init__ constructor because python doesnt include an empty array in the default one and i dont know how that works and there were like 16 errors i had to fight through to get this PR working

project/thscoreboard/replays/models.py
raviddog: 0-5 is regular, extra is 6, except in IN where it gets bloated by the A and B stages so as usual zun is inconsistent and the correct stage text should just be handled by the frontend IMO

project/thscoreboard/replays/replay_parsing.py
n-rook: Can't you just do

```
    r = ReplayInfo(
        game_ids.GameIDs.TH06,
        shots[rep_raw[6]],
        rep_raw[7],
        replay.header.score,
        datetime.strptime(replay.header.date, "%m/%d/%y"),
        stages=rep_stages
    )
```

? That's all I mean.

project/thscoreboard/replays/replay_parsing.py
n-rook: I'm worried there are bugs I'm not seeing in this code, so I'm not comfortable merging this PR until there are basic tests included. In software development, it's almost always best to add code and unit tests at the same time. git has tools to get files from other branches so it should not be too bad to put the tests in here.

project/thscoreboard/replays/models.py
n-rook: You've got to remember that not everyone working on the frontend will know what these fields correspond to in the replay file. Even if you know at a glance, nobody else will know. Even though it's annoying to have wordy comments, it's better than having me guessing at what these fields mean and adding bugs to the frontend because I don't understand them.

I do actually want to split this file up like I did `views/` because it is already too big, but I know it will cause merge conflicts for everyone when I do so I have been putting it off

project/thscoreboard/replays/models.py
n-rook: Good thing I asked, that wasn't what I would have guessed! I actually think it's a big problem if these rows represent the start for some games and the end for others. It'll be really hard to design a consistent UI if that's the case. For now could you save only stages for eosd and mof in this PR, and then send another PR later for pcb? That way we can deal with it later without holding up this.

project/thscoreboard/replays/models.py
n-rook: yeah that's the other thing, it's easy to remove restrictions later, but harder to add them (not impossible or anything, but harder)

project/thscoreboard/replays/kaitai_parsers/th06.py
n-rook: yeah I can see why you were so frustrated by the Kaitai Struct bug!

project/thscoreboard/replays/models.py
n-rook: we should think about just doing this ourselves and always making them ordered from 0, but I don't think it matters enough to do it in this pr

project/thscoreboard/replays/replay_parsing.py
raviddog: i tried that and it refused and i had enough broken things on my plate at the time

project/thscoreboard/replays/kaitai_parsers/th06.py
raviddog: #61

project/thscoreboard/replays/models.py
raviddog: addressed

raviddog: #62

project/thscoreboard/replays/replay_parsing.py
raviddog: addressed.
more tests can be added in the future, specifically I'd like to add some that involve storing and retrieving from the database.
But these changes should suffice for the current splits feature being added

project/thscoreboard/replays/kaitai_parsers/th06.py
raviddog: I've edited the .ksy file to include this, I tested and this is indeed the same code it produces with the -15 added.

project/thscoreboard/replays/models.py
raviddog: I've added this info as a comment

project/thscoreboard/replays/models.py
raviddog: addressed

project/thscoreboard/replays/replay_parsing.py
n-rook: yeah, it's great to have lots of tests, but all I think we really _need_ for this PR is basic tests that show the feature works at all

n-rook (approved):

PR #64, opened 2022-09-20
merge migration confflict

raviddog:


n-rook (approved):
how annoying, I had read "oh yeah, django totally allows for multiple migrations to be added at the same time!" and didn't realize it required something like this. oh well

PR #65, opened 2022-09-21
forgot to multiply score by 10 in th10 stage split

raviddog:


n-rook (approved):
See, this is the sort of thing that would have been caught instantly with a test :p

PR #68, opened 2022-10-08
Restyle initial upload page.

n-rook:
It still is not super pretty, but it's better than it was before:
- Actually lists all games you can upload replays for
- Upload form is still a little ugly (you can't restyle a form input without weird CSS hacks, and you can't make it look good without JS) but it is at least better
- Input only wants .rpy files by default
- Translatable


n-rook:
image

PR #70, opened 2022-10-08
Extract create_replay to new module and add a test for it.

n-rook:


n-rook:
I mostly did this so I could call this method from other places too, to more easily create replays for other tests

PR #73, opened 2022-10-15
Add a suggested Git hook that runs flake8.

n-rook:
I got tired of the CI hook telling me flake8 didn't work. This way, I will be informed before submission.

### DIFF
--- a/git-hooks/README.md
+++ b/git-hooks/README.md
@@ -1,0 +1,6 @@
+# Git hooks
+
+This directory contains suggested Git hooks for developers.
+
+These hooks don't do anything by default. To use them, copy them into
+your ./.git/hooks directory.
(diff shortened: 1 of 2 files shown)


n-rook:
I'm going to leave this open for awhile until I decide I really like having a hook here. Of course, if anyone else wants it they can always just install it from here.

raviddog:
does that nuke your push if flake8 blows up so i dont have to realise i forgot to fix it when opening the PR? the sooner you merge that the sooner we'll all get it then, im on board 👍

n-rook:
Yeah. You have to have flake8 installed though (that is, it runs "flake8" on the CLI and if you don't have it installed, it will just say an error)

n-rook:
Oh yeah, note that you have to install it yourself like it says in the md file

PR #76, opened 2022-10-15
Fix up publishing replays to disallow editing stuff that shouldn't be edited

### LINKED ISSUE #56: Improve file replay uploading page to only make appropriate fields editable
For example, difficulty should not be editable, and score should only be editable if it's counterstopped and the player is playing a counterstop game.

n-rook:
Fixes #56


n-rook:
I gave up on using a shared template for the "replay" and "upload run without a replay" pages; it just got too annoying to figure out how to style both of them simultaneously. I think this way is better.

PR #79, opened 2022-10-17
comma delimited numbers

raviddog:


project/thscoreboard/replays/templates/replays/replay_details.html
```diff
@@ -8,7 +8,8 @@ <h3>{{ difficulty_name }} ({{ route_name }}) — {{ shot_name }}</h3>
     <h3>{{ difficulty_name }} — {{ shot_name }}</h3>
     {% endif %}
     <div class="replay-box">
-        Total score: <span class="replay-points">{{ replay.points }}</span>
+        {% load humanize %}
```
n-rook, line 11: I think we should always put load tags at the top of html files. After all we wouldn't do imports in the middle of a Python file or includes in a C++ file.

n-rook (changes requested):
This is a good catch. I didn't think of adding this but it obviously benefits the site.

n-rook (approved):

PR #80, opened 2022-10-17
find & replace points to score

raviddog:
every time i see the word points in the code my brain thinks we have some kind of ranking system


n-rook (approved):
I still think "score" was a good name for a run :kyouko_pensive:

PR #85, opened 2022-10-23
Frontend stagesplits

raviddog:
opening this PR just to get comments on the code, it's not ready yet

how do i make the table_fields mess in replay_details() not awful


project/thscoreboard/shared_content/static/style.sass
```diff
@@ -143,3 +143,7 @@ th, td
             background-color: #fff3ff
         tr:nth-child(even)
             background-color: #ffe6ff
+
+.tnum
```
n-rook, line 147: I always use relatively long names for CSS classes just because it's easy to forget what they are for.

project/thscoreboard/replays/views/view_replay.py
```diff
@@ -12,20 +12,81 @@
...
+        'lives': True,
+        'life_pieces': False,
+        'bombs': False,
+        'bomb_pieces': False,
+        'th06_rank': False,
+        'th07_cherry': False,
+        'th07_cherrymax': False
+    }
+
+    game_fields = {
```
n-rook, line 69: So, yeah. This is the big, interesting problem.

Here is what I recommend:

1. Extract these big dictionaries into a separate module. Call it `game_fields` or something. I almost want to say put it in models.py, but models.py is already really really big so I don't want more stuff there.
2. Declare them as constants, and use https://pypi.org/project/immutabledict/ to make sure the constants are immutable. (I have had really insane bugs happen in the past where I declare constants, but they are mutable, and later they get modified by mistake.)
3. Add a new method, `GetTableFields()`, on the Game model, that returns the table fields for that model.
4. I would actually then write a unit test for each game (well, not the pc98 ones, obviously) that calls `PublishNewReplay()` and makes sure the expected fields are set and unset. This will be good to catch bugs and typos.
5. Now this view doesn't really have to get much more complicated! `table_fields` can just be `replay_instance.shot.game.GetTableFields()`.

I think this would basically work, and wouldn't be too bad to implement. More importantly I don't think it wouldn't be too bad to maintain either. There are probably better ways but I can't think of them.

project/thscoreboard/replays/templates/replays/replay_stages.html
```diff
@@ -0,0 +1,91 @@
+{% load humanize %}
+<table class="replay-table">
+    <thead>
+        <tr>
+            {% if table_fields.stage %}
```
n-rook, line 5: All these ifs look kinda silly and repetitive but I think they really are the way to go.

project/thscoreboard/replays/templates/replays/replay_stages.html
```diff
@@ -0,0 +1,91 @@
+{% load humanize %}
```
n-rook, line 1: Unfortunately for new code I think we really should tag it for translation when we submit it. You can take a look at `upload.html` to see how to tag strings for translation; basically you `{% load i18n %}` at the top, then you just wrap all text strings like `{% translate "Stage" %}`, or there's a `blocktranslate` too iirc.

project/thscoreboard/shared_content/static/style.sass
n-rook, line 147: nit: I think I would put a class on `tbody` instead of `tr`. All the rows are the same anyway. This is just a nitpick though, I don't think it really matters

raviddog: Right-aligned looks horrible on text

raviddog: You're right, the shorthand was a leftover from my attempt to put the class on the <td> tag, and it absolutely ballooned the generated html to the point where I started considering bandwidth

n-rook: Yeah I'm not saying you should right-align text. But the only things under `tbody` are numbers, right? All the headers are under `thead`.

raviddog: There is text on the replay scoreboard table

project/thscoreboard/replays/templates/replays/replay_stages.html
raviddog: Yeah it looks dumb but I don't see a more efficient way of doing it, plus it'll only need to be done once. It's important that we don't forget to modify it every time we add support for a new game though, and the list of files that do need to be modified when that happens is growing, which concerns me slightly. Maybe we can make a list of them so we don't forget once other people pess familiar with the codebase need to touch it

project/thscoreboard/shared_content/static/style.sass
n-rook: Yeah but if you put the class on `tbody` it won't affect the other replay table, right?

raviddog: I can do that?????
Ok lmao

n-rook: tbh I think in the long run we will have to put the classes on the `td`s anyway when we do stuff like, say, using icons for life pieces or bomb pieces... but in the short term I don't wanna do that either lol

project/thscoreboard/replays/views/view_replay.py
```diff
@@ -96,3 +157,14 @@ def GetReplayOr404(user, replay_id):
     if not replay_instance.IsVisible(user):
         raise Http404()
     return replay_instance
+
+
+def GetReplayStages(user, replay_id):
+    try:
+        replay_stages = models.ReplayStage.objects.filter(replay=replay_id)
+        replay_instance = models.Replay.objects.get(id=replay_id)
```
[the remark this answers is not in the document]
raviddog: Done

project/thscoreboard/replays/views/view_replay.py
```diff
@@ -96,3 +106,18 @@ def GetReplayOr404(user, replay_id):
...
+    try:
+        replay_instance = models.Replay.objects.select_related('shot').get(id=replay_id)
+    except models.Replay.DoesNotExist:
+        return Http404()
+    if not replay_instance.IsVisible(user):
+        return Http404()
+    try:
+        replay_stages = models.ReplayStage.objects.filter(replay=replay_id)
+    except models.Replay.DoesNotExist:
+        replay_stages = None
```
n-rook, line 122: I think this'll never happen, right? Since it never happens, you don't need to catch it. The user getting a 500 is appropriate if something that should never happen somehow does; that way they can go complain to us and we can fix it (or someday we can notice it with monitoring).

project/thscoreboard/replays/views/view_replay.py
```diff
@@ -96,3 +106,18 @@ def GetReplayOr404(user, replay_id):
...
         raise Http404()
     return replay_instance
+
+
+@transaction.atomic
+def GetReplayWithStagesOr404(user, replay_id):
+    try:
+        replay_instance = models.Replay.objects.select_related('shot').get(id=replay_id)
+    except models.Replay.DoesNotExist:
+        return Http404()
```
n-rook, line 116: I think you have to raise Http404 here; it's an exception (a special exception that Django catches and returns 404s if it sees)

n-rook (commented):
This looks basically on the right track; just add a few tests for game_fields.

project/thscoreboard/replays/views/view_replay.py
raviddog: I guess the question is what if a replay that should have stages, doesn't? Whether it's a bad parse or its an old imported royalflare run or some shit idk

project/thscoreboard/replays/views/view_replay.py
raviddog: You're right, I read the existing code wrong

project/thscoreboard/replays/views/view_replay.py
n-rook: I actually think DoesNotExist will never be raised here. Django only raises DoesNotExist if you try to get a specific object, but it does not exist. If you try to get _multiple_ objects and they don't exist, I think Django will just return the empty list.

I'm not sure what will happen if the replay itself doesn't exist (whether it raises DoesNotExist or just does nothing), but that should be impossible anyway. After all, you're in a transaction, and you already fetched the replay before. If it didn't exist, it would have raised an exception then, and would never have reached this point to begin with.

n-rook:
I'm just waiting on basic tests for game_fields; I think that's about it. Then it can go in.

FYI: You can click "Resolve conversation" if you think you've changed the code to address one of my comments. (If you think my comment is dumb, don't click "Resolve conversation"; in that case just leave a comment explaining why and if I agree, I'll resolve it myself.)

raviddog:
I don't know how to write a test for this that isn't just copy pasting the fields and checking they're the same, like that doesn't feel real-world

project/thscoreboard/replays/game_fields.py
```diff
@@ -0,0 +1,77 @@
...
+_game_fields = immutabledict({
+    'th01': None,
+    'th05': None,
+    'th06': _table_fields_th06,
+    'th07': _table_fields_th07,
+    'th10': _table_fields_th10
+})
+
+
+def GetPower(game_id: str, power: int):
```
n-rook, line 63: I only just noticed this now: This function actually returns an int and a str in different circumstances. I think it'd be better to always return a str (and thus return str(power) for th06 and th07).

n-rook:
You're right that this is hard to test. The only bugs you will be able to catch with a test here are really dumb ones. Fortunately, we all make really dumb mistakes all the time, so it is good to catch those anyway.

What I would do here is write a test. This test will call GetGameField for every game with fields. It will then compare the keys in the dictionaries returned, and will fail if any of the games have a key that isn't present in any of the other games. This way, when we add a new game, we will add it to this test, too, and it will automatically catch us if we accidentally typo a game field or forget one.

You should also write tests for GetPower. But those tests will be even dumber, you'll just call it with an old game and confirm it does nothing, and then also call it with a new game and confirm it formats the string correctly.

project/thscoreboard/replays/test_view_replay.py
```diff
@@ -0,0 +1,52 @@
...
+from replays import replay_parsing
+from replays.testing import test_case
+
+from . import game_ids
+from . import game_fields
+
+import logging
+
+
+def ReadTestFile(filename):
```
n-rook, line 12: Rather than define this here, I'd just use the function GetRaw in the test_replays module, which does the same thing.

n-rook (approved):
Nice! This looks basically good, just one more small comment

project/thscoreboard/replays/test_view_replay.py
raviddog: GetRaw was added into main after I created this branch, so I didn't know it was there lmao. Latest commit addresses this, I think I'll clean up the tests in another PR after this gets merged.

n-rook: oh whoops lol

PR #88, opened 2022-11-02
ksy update

raviddog:
added the kaitai-struct files I currently have


raviddog:
I added th11 support in another branch but figured I should do these in their own PR first

n-rook (approved):

n-rook:
I wonder if the new ks will fix the issue we had

raviddog:
nah it turns out i had an old version of the kaitai struct compiler installed, so these commits literally revert the code and then bring it back (whoops)
plus i didnt even touch the ones that actually had the issue

PR #89, opened 2022-11-05
Th11 support

raviddog:


project/thscoreboard/replays/game_fields.py
```diff
@@ -67,6 +84,8 @@ def GetPowerFormat(game_id: str, power: int) -> str:
         return str(power)
     if game_id == game_ids.GameIDs.TH10:
         return "%.2f" % (float(power) * 0.05)
+    if game_id == game_ids.GameIDs.TH11:
+        return "%.2f" % (float(power) * 0.05)
```
n-rook, line 88: Rather than repeating this line, I think it'd be better to group the conditionals like so:

```
if game_id in {game_ids.GameIDs.TH06, game_ids.GameIDs.TH07}:
  return str(power)
elif game_id in {game_ids.GameIDs.TH10, ...}:
  return ...
```

n-rook (changes requested):
Test, please!

project/thscoreboard/replays/game_fields.py
raviddog: I left them separate to make it easier to see which games have been added, but I guess yeah you can just ctrl F it

project/thscoreboard/replays/forms.py
```diff
@@ -13,7 +13,8 @@
 game_names = (
     ('th06', '東方紅魔郷 - Embodiment of Scarlet Devil'),
     ('th07', '東方妖々夢 - Perfect Cherry Blossom'),
-    ('th10', '東方風神録 - Mountain of Faith')
+    ('th10', '東方風神録 - Mountain of Faith'),
+    ('th11', '東方地霊殿 - Subterranean Animism')
```
n-rook, line 17: Optional: There's a Python trick you can use here. You can put an extra comma at the end of a tuple like this, so that the diffs look better in the future, and you don't have to remember to add another one when you add more games.

n-rook (approved):
pog

I'm not sure if you're actually done yet, so I'm not merging it myself, but if you are done you can merge it

PR #94, opened 2022-11-12
Frontend replaydata

raviddog:
- parse name from replay
- show more base replay data in frontend

lf feedback


n-rook:
image

raviddog:
oh right I forgot to exclude the year from PCB date
the game doesnt store the year, i've been saving 1900 becase the db needs one
also existing replays dont have the replay name saved right now, resubmit one to get it

project/thscoreboard/replays/templates/replays/replay_details.html
```diff
@@ -3,14 +3,39 @@
...
-    {% else %}
-    <h3>{{ difficulty_name }} — {{ shot_name }}</h3>
+    <div>Upload Date: {{ upload_date }}</div>
+    {% if replay_date %}
+    <div>Replay Date: {{ replay_date }}</div>
     {% endif %}
-    <div class="replay-box">
-        Total score: <span class="replay-score">{{ replay.score|intcomma }}</span>
-    </div>
+    <h3><b>Replay Info</b></h3>
```
n-rook, line 10: HTML nitpick: Don't use `<b>` for a whole header like this. Just use `<h3>`. If you really want to, you can style the h3 to be bold in CSS, either globally or locally (by putting a class on this h3 and using that to select it in the CSS). This makes it clearer what's going on, easier to specify a specific style, and more natural to look the same throughout the site.

lol note: when I wrote this comment the first time I found out github renders `<b>` for some reason so the comment was unreadable and everything was giant and bold

project/thscoreboard/replays/templates/replays/replay_details.html
```diff
@@ -3,14 +3,39 @@
 
 {% block content %}
     <h2><a href="/replays/{{ game_id}}">{{ game_name }}</a></h2>
-    {% if route_name %}
-    <h3>{{ difficulty_name }} ({{ route_name }}) — {{ shot_name }}</h3>
-    {% else %}
-    <h3>{{ difficulty_name }} — {{ shot_name }}</h3>
+    <div>Upload Date: {{ upload_date }}</div>
```
n-rook, line 6: I think saying "Upload date: " and "Replay date: " feels a bit programmer-y and unnatural. I'd prefer to say something like: "Uploaded on ..." and "Played on ...".

project/thscoreboard/replays/views/view_replay.py
```diff
@@ -32,6 +32,7 @@ def replay_details(request, game_id: str, replay_id: int):
         'shot_name': replay_instance.shot.GetName(),
         'difficulty_name': replay_instance.GetDifficultyName(),
         'game_id': game_id,
+        'upload_date': replay_instance.created.strftime("%d %b %Y %H:%M:%S %p"),
```
n-rook, line 35: I think we should format the dates in the Django templates, rather than doing so here in the view. That'll make it easier for us in the future. For example if we ever want to do that weird hover thing where we only show the date but show the full datetime on hover as alt text, we won't have to change the view to do that.

In general, whenever possible I try to pass data to the template, and then format that data in the template.

It looks like Django has date and time display filters: https://docs.djangoproject.com/en/4.1/ref/templates/builtins/#date

project/thscoreboard/replays/templates/replays/replay_details.html
```diff
@@ -3,14 +3,39 @@
...
+    {% if replay_date %}
+    <div>Replay Date: {{ replay_date }}</div>
     {% endif %}
-    <div class="replay-box">
-        Total score: <span class="replay-score">{{ replay.score|intcomma }}</span>
-    </div>
+    <h3><b>Replay Info</b></h3>
+    <table>
```
n-rook, line 11: I don't really love using a table here, especially since we are using a table for replays below. But it's fine. I might go try to restyle it later.

project/thscoreboard/shared_content/static/style.sass
```diff
@@ -174,3 +174,10 @@ th, td
     .num-align
         td
             text-align: right
+
+.replay-details
```
n-rook, line 178: This doesn't seem to be working.

I don't think this would look good if it was working, though; the colors would clash with the score description.

project/thscoreboard/replays/views/view_replay.py
n-rook, line 35: On another note, this formatting is too accurate! Showing people's times to the second is too much; it's information no user would want. Personally, I would just show the date, but it would be OK to show the hour and minute too.

n-rook (changes requested):
Looks mostly good, all my comments are on the HTML/CSS. I'm glad to see you got your dev env working!

project/thscoreboard/replays/views/view_replay.py
32th-System: Also:  2008年12月31日

n-rook: Oh, that's true too; we'll want to format the date based on the language. Django supports this (and we can customize it too if we don't like the defaults; see https://docs.djangoproject.com/en/4.1/topics/i18n/formatting/#creating-custom-format-files) but the best way to do that is to use the template filters.

project/thscoreboard/replays/templates/replays/replay_details.html
raviddog: I didnt use a taboe at first and it was impossible to read because the actual fields were misaligned vertically

project/thscoreboard/shared_content/static/style.sass
raviddog: Yeah, no idea what I'm doing here lmao

project/thscoreboard/replays/views/view_replay.py
raviddog: tfw copy pasted stack overflow format code

project/thscoreboard/shared_content/static/style.sass
raviddog: removed it, someone who knows frontend better can deal with it

project/thscoreboard/replays/views/view_replay.py
raviddog: the game shows hour and minute, I think we should keep that

n-rook: Yeah, this is fine for now.

n-rook:
IMO we should only allow "this data is wrong, let me edit it" in cases where we think the data might actually be wrong (like counterstop). Otherwise it's fine to just treat it as bugs. The nice thing about not letting users edit this data is we can actually fix a bug on our side by reparsing the replay.

The publish screen is a bit wonky looking rn but it's ok, we can always go fix it up later. It needs more love anyway.

n-rook (approved):

Issue #78, opened 2022-10-16
Add survival metadata to replays


n-rook:
It would be nice to have more survival metadata for replays. Let's start with:

- 1cc
- No bomb
- Miss count (optional)

32th-System:
> I know lunarcast only asks about clear for EoSD/PCB. But I don't know how

Every replay uploader except ours pulls info out of user data, though if the game's replay screen shows the clear state, then that information is also in the replay's encoded data (aka, in th10 and onwards, that information is in the encoded data, but for th08 it's only in the USER section)

n-rook:
I'm still working on this, but it's taking awhile to get it to a state I'm happy with. Hopefully I will have a PR out next weekend

32th-System:
looks like we have all of those now

n-rook:
I kinda wanted to leave this open until I had third conditions implemented (nbb, nv etc)

PR #98, opened 2022-11-12
Add a field specifying whether the player cleared or not.

n-rook:
In theory I think it's possible to get this information from some replays. I know lunarcast only asks about clear for EoSD/PCB. But I don't know how
:kyouko_pensive: #78

I was very lazy about actually displaying it on the replay details page, because I figure I'll wait for #94 to be submitted before merging this


raviddog:
we can keep this as a user filled field for the games that dont store it, but most modern games keep track of this, its either clear or the stage the run made it to. i havent looked specifically at how this is stored but we can parse it.

raviddog:
is it possible to open pull requests into your branch if i want to make changes here?

n-rook:
It technically is; you can push a branch, and then open a pull request and set the target branch to this branch (`nrook/cleared`) rather than main. But I would recommend just telling me what I should do instead.

n-rook:
Alternately I think it's OK if this goes in as-is, and then we improve it later by autodetecting clears on more games.

raviddog (approved):
yeah i think that can work for now

PR #106, opened 2022-11-14
email: determine site URL with request.META

32th-System:


n-rook:
I would love to get rid of the weird site URL code (which I think is mostly bad), but I'm worried about trusting the header here. This is just the origin that comes from the end user, right? So couldn't someone set up their local hosts file to trick us into sending emails with links to arbitrary sites?

32th-System:
I think this would happen but I'm not sure what the problem here is

n-rook:
I think it would make us look bad if someone received a forgot password email from us that linked to goatse

32th-System:
I set up a redirect rule in nginx so that any attempt to access the server using an address other than staging.silentselene.net will 301 redirect to staging.silentselene.net. That way, the server will always use staging.silentselene.net in emails

n-rook:
yeah that should be safe I think. or just 404 if it isn't staging.silentselene.net

project/thscoreboard/users/send_email.py
```diff
@@ -1,41 +1,17 @@
...
-        None
-    ))
-
-
 def SendVerificationEmail(request, u: models.UnverifiedUser):
-    full_link = _GetFullyQualifiedLink(
-        request,
-        urls.reverse('users:verify_email', kwargs={'token': u.token})
-    )
+    full_link = request.META['HTTP_ORIGIN'] + urls.reverse('users:verify_email', kwargs={'token': u.token})
```
n-rook, line 14: I found another issue, unfortunately: This only works behind nginx, so it'll cause issues when running the server locally. I think it is probably still simplest to have the server know its domain without having to look at the request. We can ditch the dumb site.domain stuff, but I think we should probably set it with an env param and have a setting in settings.py for it.

n-rook: Hmm... Or you could use a header set by nginx (maybe origin, but I almost feel like it'd be better to use a predefined static one), and default to localhost if it isn't set. That might be better and is more along the lines of what you were doing

32th-System: In my original idea I checked whether HTTP_ORIGIN was present and if it wasn't I would use HTTP_HOST. So something like
```python
site_name = request.META['HTTP_ORIGIN'] if 'HTTP_ORIGIN' in request.META else "http://" + request.META['HTTP_HOST'] + "/"
full_link = site_name + ...
```
Your idea would be
```python
full_link = settings.SITE_BASE + ...
```
where SITE_BASE is set in settings.py. I think I might prefer the second one

n-rook: yeah I just think it's simpler to understand what the site is actually doing that way

n-rook (approved):
looks good to me

PR #107, opened 2022-11-14
Add systemd unit files for automatic backups

32th-System:
For this to work you need to set the PGPASSWORD environment variable which can be done with `systemctl edit silentselene-backup.service` and typing
```
[Service]
Environment="PGPASSWORD=Password123"
```


n-rook:
Looks good, just add a comment (in the file or in a .md file somewhere) specifying that you need to set PGPASSWORD like that, since otherwise I'll forget if we ever have to move

n-rook (approved):

Issue #58, opened 2022-09-18
gettext doesn't seem to work on the GitHub CI bot


n-rook:
_('th06') returns 'EoSD' both locally and on the server, but not on GitHub.

Maybe gettext isn't installed?

32th-System:
Perhaps we need to run compilemessages in run_python_tests.yml?

n-rook:
I never figured out a good solution for this

PR #109, opened 2022-11-15
modify strings for use with translation

raviddog:
I edited the .po files with poedit, seems to work fine
Since we're using translation, the strings we write in the code don't need to be formatted the way you would for a frontend, and in fact should contain additional context information to make it clear where the string will be found

E.g. I've added the game id to every shot name string, so that you can tell what it is in the translation editor, instead of just seeing 12 lines of "ReimuA"

Also if we intentionally make the base strings look bad, it will be more obviously out of place if we've forgotten a translation


raviddog:
addresses #66

raviddog:
Not sure what to do about the test situation

raviddog:
#58 appears to be related too

raviddog:
For what its worth, the tests pass on my local machine, so this might be #58 kicking in

project/thscoreboard/replays/game_ids.py
```diff
@@ -1,6 +1,6 @@
 """Various human-readable game IDs, used in low-level libraries like game_ids.py."""
 
-from django.utils.translation import gettext as _, pgettext
+from django.utils.translation import gettext as _
```
n-rook, line 3: I think we should keep using pgettext, rather than putting context in the translation string itself. It is true that putting context in the translation string has the advantage that it makes untranslated strings look really ugly (and thus tells us we should translate them). But I don't see any benefit to translating strings that are already in English. And the downside is that it makes it more difficult to understand exactly what the strings are going to look like on the website; we have to mentally delete the context part first. gettext already has a way to add context for strings; we might as well use it rather than trying to invent our own alternative.

project/thscoreboard/replays/forms.py
```diff
@@ -26,20 +20,11 @@
...
-
 category_names = (
-    ('regular', 'Regular'),
-    ('tas', 'Tool-Assisted'),
-    ('unranked', 'Unranked'),
-    ('private', 'Private')
+    ('1', _('Regular')),
+    ('2', _('Tool-Assisted')),
+    ('3', _('Unranked')),
+    ('4', _('Private'))
```
n-rook, line 27: Why move to ints from strings for the category names? I don't think the strings were doing any harm. I mean, I don't think there's anything wrong with the ints either, but strings would be easier to read if I were looking at the request in the Chrome debugger or something like that.

README.md
```diff
@@ -77,3 +77,14 @@ score table; the games will be abbreviated in the style of "EoSD" if the
 names are translated into English, but will be in the style of "th06" otherwise.
 
 There isn't a hook yet to actually switch the language to Japanese.
+
+### Adding support for new games
```
n-rook, line 81: Nice, this is good to have written up

n-rook:
> #58 appears to be related too

Yes, I think this is unfortunately #58. For this PR I would just work around it in one hacky way or another (maybe change the assertions so they're more general and both accept the untranslated string and the translated one), because #58 seems like it will be a pain in the ass to fix and you definitely won't want to do it as part of this.

project/thscoreboard/replays/forms.py
```diff
@@ -5,18 +5,12 @@
 
 from django import forms
 from django.core import exceptions
+from django.utils.translation import gettext as _
 
 from . import game_ids
 from . import models
 from . import limits
 
-game_names = (
```
n-rook, line 13: Thanks for cleaning up this junk in this file.

n-rook (changes requested):

project/thscoreboard/replays/forms.py
raviddog: had to do that to make this fit in to replace pulling the model itself for the dropdown box, so that we could translate those strings

project/thscoreboard/replays/game_ids.py
raviddog: i was typing a reply that aaid idk what the difference between the two is, and then i realised the difference, so yeah ok ill change this

project/thscoreboard/replays/forms.py
n-rook: oh OK

n-rook (approved):
Looks good once updated from master (which I think should be easy, iirc there isn't anything that _really_ conflicts with this PR in there)

raviddog:
i already did

n-rook:
oh ok :D

PR #111, opened 2022-11-16
Don't specify HOST and PORT when connecting to db

32th-System:
This is not only not needed (at least on Windows and Linux), it also breaks the ability to connect to the database at all on the server. Developers who do need those lines can simply add them back, but they should not commit them

With this, there are no longer any modifications made locally on the server which means automatic deployment can be set up


project/thscoreboard/thscoreboard/settings.py
```diff
@@ -135,8 +135,6 @@ def _GetAllowedHosts():
     DATABASES = {
```
n-rook, line 135: I think the right thing here is probably to drop the Heroku "if" (since... we aren't using heroku, obviously) and add a different if instead for the server (or for the unit tests).

Unfortunately changing the unit tests is kind of a pain in the ass afaik, GitHub only uses the base that's actually on prod (although at least everything else comes from the PR). So if you have to fiddle with the tests to add an env variable, probably the easiest thing to do is fork the whole repo, make a bunch of commits to main trying to fix it, and then make a PR from that once you have it working

n-rook (changes requested):

n-rook (approved):

PR #112, opened 2022-11-18
uWSGI, automatic deployment, static file serving through nginx

32th-System:


ref/server/silentselene.service
```diff
@@ -5,11 +5,12 @@ After=postgresql.service
 
 [Service]
 Type=simple
+KillSignal=SIGINT
```
n-rook, line 8: Huh, what was wrong with SIGTERM?

ref/server/nginx/silentselene
```diff
@@ -7,4 +7,16 @@ server {
         location / {
                 proxy_pass http://127.0.0.1:8000;
         }
+		
+        location /static {
+                root /home/silentselene/silentselene/project/thscoreboard/shared_content;
```
n-rook, line 12: I think we will be better off if we run "compilescss" and "collectstatic", and then expose staticfiles, rather than exposing shared_content directly. That way you don't have to do stuff like refer into dist-packages. You can see the script I used for heroku deployments in `bin/post_compile`; I think that's pretty much what we would want here too (although maybe not running the commands through the shell like that).

n-rook (changes requested):
Nice, this is really great!

ref/server/silentselene.service
32th-System: the docs recommend SIGINT or SIGQUIT and say
> Sending the SIGTERM signal will obtain the same result reload-wise but will not wait for the completion of running requests.

project/thscoreboard/thscoreboard/autodeploy.py
```diff
@@ -0,0 +1,36 @@
...
+    import uwsgi
+
+    from django.core.management.commands import migrate
+    from replays.management.commands import setup_constant_tables
+    from thscoreboard import settings
+    
+    import os
+
+    def deploy(request):
+        if 'Deploy-Key' in request.headers and request.headers['Deploy-Key'] == settings.DEPLOY_KEY:
```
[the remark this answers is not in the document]
32th-System: Since HTTP basic authentication also asks for a username I decided to do authentication with user accounts. Superusers can use /deploy

ref/server/nginx/silentselene
32th-System: I implemented that though I have one question: what are your thoughts on writing CSS directly rather than compiling SASS into CSS? From what I've seen with our SASS file and the CSS it compiled to SASS is just CSS with no curly braces or semicolons so I really don't get why we don't just write CSS directly

ref/server/silentselene.service
n-rook: Ah, makes sense.

ref/server/nginx/silentselene
n-rook: The big advantage is nesting--- you can't nest CSS rules inside other rules like you can with SASS (or SCSS). The mixins are also really useful. We aren't really using the mixins much right now but we do use nesting, and I think it makes things much easier to read.

There are actually two very similar types called "SASS" and "SCSS"; SASS is the one we use, and it's where the slightly different syntax comes from. SCSS has all the same features but has regular CSS syntax.

n-rook (approved):

Issue #75, opened 2022-10-15
Automatically deploy new commits to staging


n-rook:
We should figure out how to automatically deploy new commits to staging. This is something that was really easy with Heroku but becomes more complicated when we are running on our own VPS.

n-rook:
It works! Big thanks to @32th-System for setting up most of this.

PR #114, opened 2022-11-19
Don't apply csrf for autodeploy, so webhooks can easily call it

n-rook:


n-rook:
#75

32th-System (approved):
I don't really know what any of this means  so I assume you know what you're doing

n-rook:
csrf is important because javascript is really dumb, so it's a way to stop bad websites from sending requests to good websites and thus getting user data by tricking the user's browser https://docs.djangoproject.com/en/4.1/ref/csrf/

but it doesn't really apply to calls that aren't authorized via cookies or other stuff like that, and it stopped me from just sending a POST to do the autodeploy

n-rook:
actually I'm slandering javascript. what's dumb is the standard for sending http calls through browsers. javascript is just dumb for unrelated reasons

PR #123, opened 2022-11-21
Add requirements-dev.txt.

### LINKED ISSUE #37: dev-only requirements.txt file
There are a few modules that are required at build or compile time but not at runtime. As such they shouldn't actually be deployed. We should write a separate requirements.txt including them.

n-rook:
requirements-dev.txt is a file that you can use to install both libraries used by the server and testing/local development libraries.

With the way our deployment works right now, our prod server has to know how to deploy itself, so we can't put things like the SASS compiler here. But that's OK, there's no real downside to having them be in prod.

Fixes #37

### DIFF
--- a/.github/workflows/run_python_tests.yml
+++ b/.github/workflows/run_python_tests.yml
@@ -35,8 +35,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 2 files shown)


n-rook:
We won't really know if this worked or not until it is checked in, so it might break stuff. Uh, hopefully it doesn't : D

PR #130, opened 2022-11-28
Improve replay comment handling a little bit

32th-System:
I still want to discuss designs a little more but I put everything in place and anything else should only require editing CSS

### DIFF
--- a/project/thscoreboard/replays/templates/replays/replay_details.html
+++ b/project/thscoreboard/replays/templates/replays/replay_details.html
@@ -44,6 +44,9 @@
             </tr>
         </tbody>
     </table>
+    <div class="replay-comment">
+        {{ replay.comment }}
+    </div>
     <p>
         {% if has_stages %}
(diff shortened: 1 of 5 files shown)


project/thscoreboard/replays/templates/replays/replay_details.html
```diff
@@ -44,6 +44,9 @@ <h2><a href="/replays/{{ game_id}}">{{ game_name }}</a></h2>
             </tr>
         </tbody>
     </table>
+    <div class="replay-comment">
+        {{ replay.comment }}
```
n-rook, line 48: nitpick: If something is a sentence or a paragraph, it's best to put it in a `<p>` tag, rather than using a `<div>`. That way it winds up styled like other text automatically.

n-rook (changes requested):
I'm pretty sure this is good, but could you attach screenshots to the PR so I can double-check that it doesn't look really silly?

n-rook (approved):

PR #131, opened 2022-12-01
Th08

raviddog:
I'm not sure how to add route yet, please advise
It can be autodected from the replay, I'm just not sure how to modify the replay saving functions to add it to the database


project/thscoreboard/replays/templates/replays/replay_details.html
```diff
@@ -5,7 +5,7 @@
     <h2><a href="/replays/{{ game_id}}">{{ game_name }}</a></h2>
     <div>Uploaded on: {{ replay.created|date:"d F Y, h:i A" }}</div>
     {% if replay.timestamp %}
-        {% if game_id == "th07" %}
+        {% if game_id is "th07" or "th08" %}
```
n-rook, line 8: This doesn't look right to me; does it really work properly? I would have expected you'd have to do if (game_id is "th07") or (game_id is "th08") or something like that.

project/thscoreboard/replays/replay_parsing.py
```diff
@@ -137,6 +139,59 @@ def _Parse07(rep_raw):
     return r
 
 
+def _Parse08(rep_raw):
```
n-rook, line 142: replay parsing needs tests too (although I'm sure you already know that...)

project/thscoreboard/replays/management/commands/setup_constant_tables.py
```diff
@@ -62,6 +63,30 @@ def _Create10():
         shot_row.save()
 
 
+@transaction.atomic
```
n-rook, line 66: To add routes, you should just be able to create the routes here, and then set them in the parser. (I already said this in discord, but wanted to say it here too. I'm sure you already know this.)

n-rook (changes requested):
Looks good so far; looking forward to seeing it once it's done

project/thscoreboard/replays/views/create_replay.py
```diff
@@ -96,6 +96,10 @@ def publish_replay(request, temp_replay_id):
     replay_info = replay_parsing.Parse(bytes(temp_replay.replay))
     shot_instance = models.Shot.objects.select_related('game').get(game=replay_info.game, shot_id=replay_info.shot)
 
+
+    if replay_info.game in ["th01", "th08"]:
```
n-rook, line 99: Nitpick: I think we should use game_id.GameIDs.TH08 instead of directly using the string literals. The reason I think this is that it makes typos less likely.

n-rook (approved):

PR #132, opened 2022-12-02
Display an error when user uploads bad replay

32th-System:


project/thscoreboard/replays/views/create_replay.py
```diff
@@ -61,6 +63,10 @@ def upload_file(request):
 
             except ValidationError as e:
                 form.add_error('replay_file', e)
+            except MemoryError as e:
```
n-rook, line 66: It's generally considered not a good idea to catch MemoryError.

The reason for this is that if Python raises MemoryError, something very weird is going on in the server. It should never happen under normal circumstances, no matter what is uploaded (there's a reason we limit the size of replays, after all). As such, we should let the exception bubble up to Django. That way, it will result in a 500 error. When we add internal monitoring later (with something like Datadog or Honeycomb, or built ourselves), we can track all 500s and know there is a bug when one happens.

(At some point, we should add a generic 500 error page, as mentioned in https://docs.djangoproject.com/en/4.1/ref/views/#error-views.)

project/thscoreboard/replays/views/create_replay.py
```diff
@@ -61,6 +63,10 @@ def upload_file(request):
...
+                form.add_error('replay_file', _("The server could not handle your request at this time, please try again later") + f" (MemoryError: {e})")
+            except Exception:
```
n-rook, line 68: Similar to the above, it's not a best practice in Python to catch Exception unless writing a top-level "unexpected error" handler or reraising it. The reason is that all sorts of things can be Exceptions. Bugs especially result in Exceptions; if we have a logic error in our replay parsing handler, it could easily return such a thing.

It should be fine just to catch the expected Errors (that is, the Error class) from the replay parser, which is what we're doing now. If bad replays can raise other errors, we should handle them in replay_parsing.py and explicitly raise them as the Error class.

n-rook (changes requested):

n-rook (changes requested):
This CL makes sense to me, but we should add tests with corrupted replays, even if they're just "we took the first 1k bytes of an eosd replay and replaced the rest with lesanae"

n-rook (approved):
Looks good!

PR #134, opened 2022-12-08
Staff features

32th-System:
This includes the ability for staff to delete other people's replays and a new feature for users to edit the comment on their replays (and for staff to edit the comment on anyone's replays)

### DIFF
--- a/project/thscoreboard/replays/views/view_replay.py
+++ b/project/thscoreboard/replays/views/view_replay.py
@@ -8,7 +8,20 @@ from django.db import transaction
 
 from replays import models
 from replays.lib import http_util
+from replays import forms
 from replays import game_fields
+
+
+@http_decorators.require_POST
+def edit_replay(request, game_id: str, replay_id: int):
+    replay_instance, replay_stages = GetReplayWithStagesOr404(request.user, replay_id)
+
+    edit = forms.EditReplayForm(request.POST)
+    if edit.is_valid() and replay_instance.user == request.user:
+        replay_instance.comment = edit.cleaned_data['comment']
+        replay_instance.save()
+
+    return redirect(replay_details, game_id=game_id, replay_id=replay_id)
 
 
 @http_decorators.require_safe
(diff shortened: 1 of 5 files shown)


project/thscoreboard/replays/views/view_replay.py
```diff
@@ -8,13 +8,28 @@
...
 def replay_details(request, game_id: str, replay_id: int):
     replay_instance, replay_stages = GetReplayWithStagesOr404(request.user, replay_id)
 
+    edit_comment = forms.EditReplayCommentForm(
+        initial={
+            'comment': replay_instance.comment
+        }
+    )
+
+    if request.method == 'POST':
```
n-rook, line 25: I think this should go to a different URL, rather than the POST happening on the same URL. After all, someday maybe we'll have multiple of these edit options, and it'd be awkward to handle them all in the same place. This can be controlled by https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-action.

project/thscoreboard/replays/views/view_replay.py
```diff
@@ -34,11 +49,12 @@ def replay_details(request, game_id: str, replay_id: int):
         'difficulty_name': replay_instance.GetDifficultyName(),
         'game_id': game_id,
         'replay': replay_instance,
-        'is_owner': request.user == replay_instance.user,
+        'is_owner': request.user == replay_instance.user or request.user.is_staff,
```
n-rook, line 52: I think this is a bit confusing; I would expect "is_owner" to be set only if the user is the person who uploaded the replay. I'd rather see us add a new variable, called something like "can_edit" or something.

project/thscoreboard/replays/views/view_replay.py
```diff
@@ -8,13 +8,28 @@
...
 
+    edit_comment = forms.EditReplayCommentForm(
+        initial={
+            'comment': replay_instance.comment
+        }
+    )
+
+    if request.method == 'POST':
+        edit_comment = forms.EditReplayCommentForm(request.POST)
+        if edit_comment.is_valid() and (replay_instance.user == request.user or request.user.is_staff):
```
n-rook, line 27: So... I don't think we should let mods edit comments, and here's why: The comment is something that, on the website, we're claiming was said by the user. As such, I worry that if we let mods edit comments, we are basically putting words in someone else's mouth. Maybe we will want it at some point, but it's the sort of feature which I think we need to be really careful about how we add. At the very least, we'd need some sort of notice so users know the comment was edited by a mod.

project/thscoreboard/shared_content/static/js/replays/replay_details.js
```diff
@@ -0,0 +1,11 @@
+window.onload = (event) => {
+    let el = document.getElementById("replay-comment-edit");
```
n-rook, line 2: Rather than setting display = "none" on load, can we just set display = "none" in the CSS?

n-rook (changes requested):
Very neat!

First JavaScript 😲

project/thscoreboard/shared_content/static/js/replays/replay_details.js
32th-System: I did this so that if JavaScript is disabled the edit box is still available

n-rook: oh that makes sense. can you add a comment here saying that?

n-rook (approved):
\o/

PR #135, opened 2022-12-11
Th09

raviddog:
The replay details is sort of a mess, and doesn't really say much. But the replays themselves don't have that much info anyway. Importantly, final score is not saved, and so we might actually for the first time need to let users enter their own score and have rep_score fields be different from the actual score (and show it on the frontend).


raviddog:
I would prefer to merge this now, and create an issue for adjusting the frontend + proper score submission, if that's fine with people

project/thscoreboard/replays/create_replay.py
```diff
@@ -73,6 +74,11 @@ def PublishNewReplay(
     temp_replay_instance.delete()
 
     for s in replay_info.stages:
+        #   th09 shot foreign key
+        shot_instance = None
```
n-rook, line 78: nitpick: IMO this part of the code would be clearer if you first created the ReplayStage instance, and then did `if replay_info.game == game_ids.GameIDs.TH09:` and set stuff on the instance then. like,

```
replay_stage = ...

if replay_info.game == game_ids.GameIDs.TH09:
  shot_instance = ...
  etc

replay_stage.save()
```

project/thscoreboard/replays/views/create_replay.py
```diff
@@ -95,6 +95,10 @@ def publish_replay(request, temp_replay_id):
         raise Http404()
 
     replay_info = replay_parsing.Parse(bytes(temp_replay.replay))
```
n-rook, line 97: I totally get why you want to merge this PR; trying to do additional work on top of this would get really unwieldy really fast. However, I don't think this is quite ready for users yet.

As such, I think probably the right thing to do is actually to use the DEBUG setting and not let people upload th09 replays outside of debug. You can import DEBUG from https://docs.djangoproject.com/en/4.1/topics/settings/#using-settings-in-python-code, and then raise an error if the replay is PoFV. It'd probably be best to do this under upload_file().

Of course... it does not really matter if it is ready for users yet, since we have not released. But I figure it is good to get in the habit, and what I'm suggesting should be a easy change. This way we will know what to do when adding new games later.

n-rook (changes requested):

project/thscoreboard/replays/create_replay.py
raviddog: I'm not sure about that, I think when i create the ReplayStage instance, I want to have every single field in there rather than having them spread out. It makes it easier to notice when a field has been missed.

project/thscoreboard/replays/views/create_replay.py
raviddog: I think the actual replay parsing and uploading is all fine, it's just that the frontend is a little bit ugly, which I'm not sure we care about that much right now. I was just mainly concerned that editing the form would be a lot of work, but I'm managed to do it now

n-rook: 👍

project/thscoreboard/replays/create_replay.py
n-rook: If you're happier with it this way that's fine too.

n-rook (approved):

Issue #92, opened 2022-11-06
Self-host web fonts


n-rook:
We are currently using web fonts from Google's font delivery service. Web fonts are nice but some folks don't like Google, so we should probably self-host these fonts so Google doesn't have to be involved.

32th-System:
The actual creators of the fonts we use host them here:
- [Lato](https://www.latofonts.com/lato-free-fonts/)
- [EB Garamond](https://github.com/octaviopardo/EBGaramond12/tree/master/fonts/webfonts)

n-rook:
Oh nice!

PR #141, opened 2022-12-12
Self host EB Garamond as well

32th-System:
It makes the font look slightly different now but I think I prefer it and it also makes it looks sharper


project/thscoreboard/shared_content/static/fonts/fonts.css
```diff
@@ -0,0 +1,201 @@
...
+}
+
+@font-face {
+    font-family: 'Lato';
+    src: url('Lato-BlackItalic.woff2') format('woff2');
+    font-style: italic;
+    font-weight: 900;
+}
+
+
```
n-rook, line 131: I think we should keep the two fonts in different files, just to make it easier to keep them straight.

(Actually, I just found out about "SASS modules" https://sass-lang.com/guide; we should definitely use those at some point so that all our files get compiled into one place. But no need to do that in this PR.)

project/thscoreboard/shared_content/static/fonts/fonts.css
```diff
@@ -0,0 +1,201 @@
...
+@font-face {
+    font-family: 'EB Garamond';
+    src: url('EBGaramond-Bold.woff2') format('woff2');
```
n-rook, line 134: wtf there are two EB Garamonds

I had a whole comment written out going "where did you find these, the garamond I found is only two files??" before I figured it out

(no action required)

In the medium term, since this version has real bolds, we should stop using meme font weights like 550 and use only the widths defined here

project/thscoreboard/replays/templates/replays/docs/licenses.html
```diff
@@ -10,7 +10,7 @@ <h2>This Site</h2>
     </p>
     <h2>Other Components</h2>
     <p>
-        We use <a href="https://www.latofonts.com/">Lato</a>, a sans serif font, on Silent Selene. Lato was created by Łukasz Dziedzic, and is licensed under the <a href="http://scripts.sil.org/OFL">SIL Open Font License 1.1</a>.
+        We use the fonts <a href="https://www.latofonts.com/">Lato</a> and <a href="https://github.com/octaviopardo/EBGaramond12">EB Garamond 12</a>, both of which are licensed under the <a href="http://scripts.sil.org/OFL">SIL Open Font License 1.1</a>.
```
n-rook, line 13: I think we should keep explicitly giving credit to the font creators in this file. It's just a nice thing to do.

Actually, I'd like to add a credits page so we give credit to ourselves too. Someday...

project/thscoreboard/shared_content/static/fonts/fonts.css
```diff
@@ -0,0 +1,201 @@
...
+
+@font-face {
+    font-family: 'Lato';
+    src: url('Lato-BlackItalic.woff2') format('woff2');
+    font-style: italic;
+    font-weight: 900;
+}
+
+
+@font-face {
```
n-rook, line 132: Please sort the @font-face tags by weight, so it's easy to check and make sure we didn't forget any.

n-rook (changes requested):
Looks good, thanks!

In the future, if there's an assigned GitHub issue open for something (like #92), please try to ask the person to whom its assigned before working on it. In this case it was fine, I hadn't done anything on EB Garamond so no effort was wasted. But I would have been sad if I had a PR that was 90% done.

This is part of why I try to file issues; it helps us not forget work we have to do, but it also helps prevent us from all writing the same code and accidentally wasting effort

project/thscoreboard/replays/templates/replays/docs/licenses.html
32th-System: It doesn't have to be someday

n-rook: You didn't put back the Lato guy's name! That's what I meant when I said "keep explicitly giving credit to the font creators".

n-rook (changes requested):
thanks, this is basically ready

project/thscoreboard/replays/templates/replays/docs/licenses.html
32th-System: It's in credits.html

n-rook (approved):

PR #143, opened 2022-12-13
th07/th08/th09: don't pass extra data into LZSS

32th-System:
right now _Parse08 and _Parse09 will send the compressed data and the USER section into the decryption and then the decompression function. This leads to some replays causing an LZSS error


n-rook (changes requested):
Add regression tests! For a bug fix like this, we should add tests with the bug fix, so we don't accidentally reintroduce the bug while refactoring the replay parsing code later.

PR #148, opened 2022-12-16
support spell practice

raviddog:
add detection and ui (spell card id) for spell practice replays (currently only TH08)


project/thscoreboard/replays/templates/replays/replay_details.html
```diff
@@ -46,6 +46,12 @@ <h2><a href="/replays/{{ game_id}}">{{ game_name }}</a></h2>
                 <td>Clears?</td>
                 <td>{% if replay.is_clear %}Yes{% else %}No{% endif %}</td>
             </tr>
+            {% if replay.spell_card_id %}
```
n-rook, line 49: I think in addition to this, we should also show the spell card ID on the "publish replay" page that shows up when you have already uploaded the replay but get the chance to add a comment or something like that.

n-rook, line 49: I think we should change some of the lists too, if we are going to include spell practice replays. Some of the lists should not have spell practice. Other lists should, but they should be marked so that they're easily distinguished by run. I think the difficulty page (like https://www.silentselene.net/replays/th06/d2) and the difficulty+shot page (https://www.silentselene.net/replays/th06/d2/ReimuA) should probably not have spell practice replays, since they really should be functioning as scoreboards. But they should definitely show up on the front page since that's telling people about every new upload.

n-rook (changes requested):
Looks good, but I have a few pain in the ass comments for extra stuff we should do before letting people upload spell practice replays

project/thscoreboard/replays/templates/replays/replay_details.html
raviddog: I think we'll want to reorganise these pages at some point in the future, make it easier to navigate, etc
But for now, I've excluded non-Regular (detected as full run) replays from those pages

project/thscoreboard/replays/models.py
```diff
@@ -214,6 +214,9 @@ def GetDifficultyUrlCode(self):
     but technically there are 12 bytes allocated, so space is reserved just in case
     """
 
+    spell_card_id = models.IntegerField(blank=True, null=True)
+    """In the case of a spell practice replay, the spell card ID attempted"""
+
```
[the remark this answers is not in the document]
raviddog: I think I've done this. I added a spell practice replay to the db saving tests, I'm not sure how to test violating the constraints though

project/thscoreboard/replays/views/create_replay.py
```diff
@@ -140,11 +142,15 @@ def publish_replay(request, temp_replay_id):
         'route_name': None,
         'has_replay_file': True,
         'replay': replay_info,
+        'replayType': game_ids.GetReplayType(replay_info.replayType)
```
n-rook, line 145: nitpick: let's call this replay_type, not replayType, for consistency with the other context variables

project/thscoreboard/replays/models.py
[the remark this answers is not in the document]
n-rook: I think that should work, yeah. For a test, you can add something to test_models that creates a spell practice replay, and then try to set its replay_type field to normal and save it, and then check that this raises an exception.

project/thscoreboard/replays/views/replay_list.py
```diff
@@ -15,9 +15,10 @@ def game_scoreboard(request, game_id: str, difficulty: Optional[int] = None, sho
     # Ancient wisdom: You don't need pagination if you don't have users yet!
     game = get_object_or_404(models.Game, game_id=game_id)
     all_replays = (
-        models.Replay.objects.select_related('shot', 'replayfile')
+        models.Replay.objects.select_related('shot')
         .filter(category=models.Category.REGULAR)
         .filter(shot__game=game_id)
+        .filter(replay_type=1)
```
n-rook, line 21: We should try to avoid magic numbers in regular code where we can; there is no way we're going to remember what 1, 2, and 3 are all the time. I think you should be able to do something like `models.ReplayType.REGULAR`. I'm not sure if there's a good way to do it in places where you don't really want to be importing `models`, so I didn't leave any comments there.

project/thscoreboard/replays/views/view_replay.py
```diff
@@ -57,13 +59,15 @@ def replay_details(request, game_id: str, replay_id: int):
...
         'replay': replay_instance,
+        'spell_name': spell_cards.get_spell_name('en', game_id, replay_instance.spell_card_id, replay_instance.difficulty),
         'can_edit': request.user == replay_instance.user,
         'can_delete': request.user == replay_instance.user or request.user.is_staff,
         'replay_file_is_good': replay_instance.is_good,
         'has_stages': len(replay_stages) != 0,
         'replay_stages': replay_stages,
         'table_fields': game_fields.GetGameField(game_id),
         'edit_form': edit_form,
+        'replayType': game_ids.GetReplayType(replay_instance.replay_type)
```
n-rook, line 70: as above, we should prefer to camel-case this

project/thscoreboard/replays/views/view_replay.py
```diff
@@ -57,13 +59,15 @@ def replay_details(request, game_id: str, replay_id: int):
         'difficulty_name': replay_instance.GetDifficultyName(),
         'game_id': game_id,
         'replay': replay_instance,
+        'spell_name': spell_cards.get_spell_name('en', game_id, replay_instance.spell_card_id, replay_instance.difficulty),
```
n-rook, line 62: We should add a function on the replay model itself to do this, kind of like GetDifficultyName(). If a function has no arguments (which should be possible if we use [get_language](https://docs.djangoproject.com/en/4.1/ref/utils/#django.utils.translation.get_language)) we don't even need to call it here, we can just pass replay_instance in and call it from the template.

n-rook (changes requested):

project/thscoreboard/replays/views/view_replay.py
raviddog: its hard because the naming standard in the DB is with underscores, but the naming standard in the dataclass is camelcase, so I'm not sure how to handle this

raviddog: oh nvm I see what you're saying

32th-System (changes requested):
Spell IDs in spells.js aka spells/{lang}/{game_id},json and in all the code I wrote around it are also 0 indexed

raviddog:
I have created another branch retaining the spell card name changes
https://github.com/raviddog/thscoreboard/tree/spell-practice-name-backup

project/thscoreboard/replays/management/commands/download_thpatch_spells.py
```diff
@@ -0,0 +1,18 @@
+from django.core.management.base import BaseCommand
```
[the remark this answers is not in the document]
32th-System: On the one hand, deleting this file would defeat the whole point of the spell name code, on the other hand, we weren't supposed to implement spell names in the same PR. I'll get rid of it and the Japanese th08.json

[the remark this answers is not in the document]
n-rook: yeah I'm not saying there's anything wrong with the file or that it shouldn't go in later, but it only makes sense with the other spell name stuff we are doing in the next PR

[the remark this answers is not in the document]
32th-System: Looks like it's all gone now

project/thscoreboard/replays/views/view_replay.py
```diff
@@ -51,6 +51,9 @@ def replay_details(request, game_id: str, replay_id: int):
         }
     )
 
+    if replay_instance.spell_card_id:
+        replay_instance.spell_card_id += 1
```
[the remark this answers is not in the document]
raviddog: Added to both the model and the Replay object in ReplayParsing.py, although the naming conventions of both are different, so the calls in the create_replay template and the replay_details template are different. Hopefully this is okay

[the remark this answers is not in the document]
n-rook: eh this is the other benefit of a linter. if flake8 doesn't complain about something like that, I won't either

n-rook (approved):

32th-System (approved):

PR #149, opened 2022-12-18
Dark theme

32th-System:


project/thscoreboard/thscoreboard/templates/base.html
```diff
@@ -41,9 +41,13 @@ <h1>Silent Selene</h1>
             <a href="/users/register">Sign up</a>
         </div>
         {% endif %}
+        <div id="topbar-link-button">
+            <a href="#" onclick="toggleTheme()">🌙</a>
```
n-rook, line 45: This button looks a bit weird with the underline. It looks like this in my browser:

image

We should style this so that the underline doesn't show up.

project/thscoreboard/shared_content/static/js/main.js
```diff
@@ -0,0 +1,13 @@
+if(localStorage.getItem("darktheme")) {
+    document.body.classList.add("darktheme");
```
n-rook, line 2: I think we should have a lighttheme class too. That way we can try to remember to only set colors and stuff for lighttheme and darktheme, rather than always setting the colors with lighttheme and then overriding them with darktheme. You don't need to go update all the colors in this PR to use lighttheme, but I think having the class around will be convenient for later.

project/thscoreboard/shared_content/static/js/main.js
```diff
@@ -0,0 +1,13 @@
+if(localStorage.getItem("darktheme")) {
```
n-rook, line 1: nitpick: let's call it `dark-theme` rather than `darktheme`

project/thscoreboard/shared_content/static/style.sass
```diff
@@ -181,3 +185,26 @@ th, td
 
 .replay-comment
     margin: 2em
+
+.darktheme
+    background-color: black
+    color: white
+
+.darktheme .replay-table
```
n-rook, line 193: We should try to put these dark theme things as close as possible to the original statement. I know in some places we can't get that close because of the way the original selectors were written, but let's at least get as close as possible.

n-rook (changes requested):
This is really nice!

project/thscoreboard/thscoreboard/templates/base.html
32th-System: I used the emoji as a placeholder and would prefer an svg instead

project/thscoreboard/shared_content/static/style.sass
32th-System: I would prefer to have the theming stuff in one place rather than the theming stuff next to the things being themed. I guess I'll add a lighttheme class for that

project/thscoreboard/thscoreboard/templates/base.html
n-rook: I agree with that too, but I still want the underline gone for now. esp since realistically we are likely to not draw those SVGs for awhile

n-rook (approved):
Nice

PR #161, opened 2022-12-21
Add a "request invite" page and hide the signup page from the site.

n-rook:
image

The link says "Get an invite". I wasn't sure what else to say on it


raviddog:
Use your branch name lmao "request an invite"
The wording better implies that we don't have to honor it

PR #163, opened 2022-12-22
Consistent stage splits

### LINKED ISSUE #60: th07 stores stage splits for end-of-stage, everything else is the start of the stage

raviddog:
fixes #60


project/thscoreboard/replays/models.py
```diff
@@ -319,7 +319,7 @@ class Meta:
                     all in all there are 40 offsets saved for potential stage data, most unused
     """
 
-    score = models.BigIntegerField()
+    score = models.BigIntegerField(blank=True, null=True)
```
n-rook, line 322: Add a note explaining why this is not set sometimes. After all, we thought it would be set all the time, so it would not be too surprising if we forgot about it later.

project/thscoreboard/replays/replay_parsing.py
```diff
@@ -98,6 +98,20 @@ def _Parse06(rep_raw):
             rep_stages.append(s)
         i += 1
 
+    """adjust stage data to be end-of-stage"""
```
n-rook, line 101: Python nitpick: `""" """` should only be used for _docstrings_; that is, strings put immediately after a function, class, or field and used as documentation. This is a regular code comment, so you should use `#`.

project/thscoreboard/replays/replay_parsing.py
```diff
@@ -147,6 +161,26 @@ def _Parse07(rep_raw):
             rep_stages.append(s)
         i += 1
 
+    for i in range(len(rep_stages)):
```
n-rook, line 164: I think this two-step process is more confusing than it needs to be. We first record the stage values wrong (above), and then we fix them here. Instead, I think it would be clearer to get them right the first time. To be honest, this code is always going to be a bit confusing, because what ZUN did makes no sense, but I think that approach would be clearer.

This would probably look something like:
```
for i, this_stage in enumerate(replay.stages):
  if replay.file_header.stage_offsets[i] != 0:
    if i + 1 >= len(replay.stages):
      next_stage = None
    else:
      next_stage = replay.stages[i + 1]
    
    s = ReplayStage()
    s.stage = i
    s.score = score.score * 10
    # …
    if next_stage:
      s.power = next_stage.power
      # ...
    else:
      s.power = None
      # ...
```

If you hate this idea, let me know, but I really do think it would be clearer. This applies to the other bug fixes too.

project/thscoreboard/replays/game_fields.py
```diff
@@ -157,6 +159,8 @@ def GetFormatPower(game_id: str, power: int) -> str:
 
 
 def GetFormatLives(game_id: str, lives: int, life_pieces: int) -> str:
```
n-rook, line 161: Fix this function (and the other ones) so that they ask for Optional[int], not int.

If you're wondering why the heck these annotations exist at all, there actually are Python type checkers, I just haven't figured out how to set one up yet.

project/thscoreboard/replays/game_fields.py
```diff
@@ -194,3 +200,48 @@ def GetFormatStage(game_id: str, stage: int) -> str:
         return str(stage + 1)
     else:
         return str(stage)
+
+
+def FormatStages(game_id: str, replay_stages):
```
n-rook, line 205: If I saw this function get called, it would be very surprising to me that it actually modified the replay_stages argument! In general I don't think we should ever modify database model instances unless we actually plan on saving those changes to the database. Otherwise it just gets too confusing to figure out what is going on with an object, since its values and behavior will be totally different from what the documentation on the model says.

In this case we have a few alternatives. I don't love any of them but I think they would all work.

1. Use [default_if_none](https://stackoverflow.com/questions/16459890/django-template-display-item-value-or-empty-string) in the template. In some sense this is "good style", since we are deciding how to display the empty value in the template itself. But it seems very repetitive. I think I would actually do this if I were writing the PR but I can see why you wouldn't want to.
2. Define a new dataclass here, called something like "DisplayStage". Instead of modifying replay_stages, create a DisplayStage for each stage and return that list instead.
3. Add lots of properties on the model; "display_power", "display_score" etc.

Let me know if you hate all of these and I will try to think of something else… don't quit the project :lesanae:

n-rook (changes requested):
Thanks for sending this so quickly!

project/thscoreboard/replays/replay_parsing.py
raviddog: I tried doing that, but it looked horrible. I had to add way more bounds checks and basically ended up parsing each stage twice. I reckon this way is clearer because we separate out replay parsing bugs from the correction code, so it's more explicit what's happening.

project/thscoreboard/replays/game_fields.py
raviddog: I'm trying to think of a situation where life_pieces is not provided and yet we still want to use this function. A ReplayStage object will always have this field, even if its set to None, so we just need a None check on the value. If there is an object that doesn't have the life_pieces field, why would we need to call this function?

project/thscoreboard/replays/game_fields.py
raviddog: My bad, it's a remnant from when I tried to use a loop to automate checking each field, but both the ReplayStage object and the model object didn't support that syntax so I ended up assigning them manually and forgot to apply the changes to a separate object instance.

project/thscoreboard/replays/replay_parsing.py
raviddog: Also some games we add later will actually contain the end stage data, so I feel that it will be confusing if some of the parsers look like a nightmare and some don't.

raviddog: The advantage of this strategy is that all the values we parse are right next to each other, so it's very easy to see if we miss anything. I value that over shortening the code.

project/thscoreboard/replays/game_fields.py
n-rook: The reason I asked for `Optional[int]` is because you edited these functions to support `None`, like

```
if lives is None:
  return ""
```

I don't really have an opinion on whether they should accept None or not, but if they do, you should annotate the types correctly. And if they don't, we should delete that code.

project/thscoreboard/replays/replay_parsing.py
n-rook: OK, if you've tried it and think it looks horrible, I'll believe you

n-rook, line 164: This would be closer to idiomatic Python if you did it like:

```
for i, this_stage in enumerate(rep_stages):
  if i < len(rep_stages - 1):
    next_stage = rep_stages[i]
    …
  else:
    …
```

rather than accessing rep_stages[i] and rep_stages[i+1] all the time. Please make that change everywhere else, too.

project/thscoreboard/replays/test_view_replay.py
```diff
@@ -26,21 +22,6 @@ def setUp(self):
         super().setUp()
         self.user = self.createUser('view-replay-user')
 
-    def testFields(self):
```
n-rook, line 29: Hmm… Do we really want to get rid of this? I mean, this was for making sure we didn't forget to set game fields for various games, right?

Can we just edit this to not check the last stage for a replay, instead?

n-rook (changes requested):

project/thscoreboard/replays/game_fields.py
raviddog: Does Optional mean it doesn't have to be that type, or does it mean the argument isnt necessary, because I was under the impression it was the latter, and I don't see us calling this function without all 3 arguments

raviddog: I'll do it anyway if you think that's best

n-rook (approved):

PR #164, opened 2022-12-22
Show spell names for spell practice replays

32th-System:


project/thscoreboard/replays/spell_cards.py
```diff
@@ -0,0 +1,36 @@
...
+    spell_names = dict()
+    for lang in thcrap_langs:
+        spell_names[lang] = dict()
+        for game_id in has_thcrap:
+            try:
+                with open(f'replays/spells/{lang}/{game_id}.json', 'rb') as f:
+                    spell_names[lang][game_id] = json.loads(f.read())
+                    logging.info(f'replays/spells/{lang}/{game_id}.json loaded!')
+            except FileNotFoundError:
+                pass
```
n-rook, line 22: These files should always be here, right? Should we log a warning if a file is missing?

project/thscoreboard/replays/spell_cards.py
```diff
@@ -0,0 +1,36 @@
...
+    return spell_names
+
+
+_SPELL_NAMES = immutabledict(_ReadSpellNames())
+
+
+def get_spell_name(lang: str, game_id: str, spell_id: int, difficulty: int = 0):
+    """Retrieves a spell card's name based on a language, the gameid of the game it's from, a 0 indexed spell ID, and a difficulty ID"""
```
n-rook, line 30: Nitpick: Don't write a really long one line docstring; try to keep them under 80 characters. Instead do this on multiple lines. See https://peps.python.org/pep-0257/.

For this one, you are really just describing the arguments. So I would do something like this:

```
"""Retrieves a spell card's name.

Args:
  lang: A language code for the name.
  game_id: The game the spell card came from.
  spell_id: The spell card's ID (0-indexed).
  difficulty: The difficulty for the game.
"""
```

This is the Google argument documentation format (see https://google.github.io/styleguide/pyguide.html#383-functions-and-methods). I don't really care about requiring docs have this "Args:" section, but I just used it because it's convenient for this case and has all the information you put in the docstring anyway.

project/thscoreboard/replays/spell_cards.py
```diff
@@ -0,0 +1,36 @@
...
+    for i in range(difficulty + 1):
```
n-rook, line 31: I don't understand this for statement; what's going on here?

project/thscoreboard/replays/spell_cards.py
```diff
@@ -0,0 +1,36 @@
+from replays import game_ids
```
n-rook, line 1: This module needs tests!

project/thscoreboard/replays/spell_cards.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def get_spell_name(lang: str, game_id: str, spell_id: int, difficulty: int = 0):
+    """Retrieves a spell card's name based on a language, the gameid of the game it's from, a 0 indexed spell ID, and a difficulty ID"""
+    for i in range(difficulty + 1):
+        try:
+            return _SPELL_NAMES[lang][game_id][f'{spell_id - i}']
+        except KeyError:
+            pass
+    return None
```
n-rook, line 36: Should we raise an exception if a spell name is not found?

project/thscoreboard/replays/management/commands/download_thpatch_spells.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Command(BaseCommand):
+    
+    help = 'Download spell names from thpatch.net'
+
+    def handle(self, *args, **kwargs):
+
+        for lang in replays.spell_cards.thcrap_langs:
+            for game in replays.spell_cards.has_thcrap:
+                os.makedirs(f'replays/spells/{lang}', exist_ok=True)
```
n-rook, line 16: I think we should try to avoid putting generated, non-checked in files into the code directories like `replays/`. Especially, if we put the jp file (which is checked in) next to the en one (which is not checked in), it will get confusing.

Honestly it might just be best to check the en file in too (although we would still want this command obviously). That way it is easy to update without server access, and jp and en are handled in the same way.

project/thscoreboard/replays/templates/replays/publish.html
```diff
@@ -36,7 +36,12 @@ <h2>{{ game_name }}</h2>
                 <td>{{ route_name }}</td>
             </tr>
             {% endif %}
-            {% if replay.spell_card_id %}
+            {% if spell_name %}
+            <tr>
+                <td>Spell Name</td>
+                <td>{{ spell_name }}</td>
+            </tr>
+            {% elif replay.spell_card_id %}
```
n-rook, line 44: IMO we can cut this part now. There should always be a spell name, right? Same with the replay_details file below.

n-rook (changes requested):
Nice!

project/thscoreboard/replays/spell_cards.py
32th-System: it's to make sure `i` reaches `difficulty` because I want to check all IDs from `spell_id` to `spell_id - difficulty`

project/thscoreboard/replays/spell_cards.py
32th-System: I'm more in favour of just displaying the ID on the site

project/thscoreboard/replays/templates/replays/publish.html
32th-System: If there is no spell name I still want to indicate that the replay is spell practice and show the spell ID as a backup

project/thscoreboard/replays/test_spell_name.py
```diff
@@ -0,0 +1,20 @@
+import unittest
```
n-rook, line 1: For tests, I think it's best to always name them test_the_name_of_the_module_they_are_testing, so it is easy to find them. For example this file should be test_spell_cards.

(Actually I think the best way would really be to call them like "spell_cards_test", so they would appear next to the module being tested when the directory is sorted alphabetically, but unfortunately the default Django test finder doesn't support that format. Maybe I'll mess around with it someday)

project/thscoreboard/replays/spell_cards.py
n-rook: Could you add a comment to that effect?

project/thscoreboard/replays/spell_cards.py
n-rook: IMO it'd be better to return 500— I'm a strong believer in [offensive programming](https://wiki.c2.com/?OffensiveProgramming)— but I won't insist

n-rook (approved):

PR #168, opened 2022-12-25
Add IP bans with subnet support

32th-System:


project/thscoreboard/users/other_views.py
```diff
@@ -47,8 +48,11 @@ def register(request):
     if request.method == 'POST':
         try:
             form = RegisterForm(request.POST)
-
-            if form.is_valid():
+            request_ip = ip_address(request.META.get('REMOTE_ADDR'))
```
n-rook, line 51: Rather than checking from within register(), I think it would be better to create a decorator that rejects a request if the user is IP banned. That way, it will be easy to reuse the same code for other routes, and it allows us to keep the view functions simpler. The downside is that this approach makes the error uglier, but I think that's fine.

This is conceptually similar to the [builtin decorators](https://docs.djangoproject.com/en/4.1/topics/http/decorators/), so you can take a look at them if you are not sure how to implement it.

project/thscoreboard/users/models.py
```diff
@@ -274,3 +275,14 @@ def RecordVisit(cls, user: Optional[settings.AUTH_USER_MODEL], ip: str):
...
+
+
+class IPBan(models.Model):
+    """Keeps a list of IP subnets of IPs banned from signup"""
+
+    ip = models.TextField()
+    """The IP address banned from signups"""
+
+    def save(self):
+        ipaddress.ip_network(self.ip)
```
n-rook, line 287: I am assuming you are doing this to make sure the IP addresses are valid, right? If that's the case, I think it would be better to implement it as a [validator](https://docs.djangoproject.com/en/4.1/ref/validators/).

project/thscoreboard/users/models.py
```diff
@@ -274,3 +275,14 @@ def RecordVisit(cls, user: Optional[settings.AUTH_USER_MODEL], ip: str):
...
             new_visit = cls(user=user, ip=ip, created=now)
             new_visit.save()
+
+
+class IPBan(models.Model):
+    """Keeps a list of IP subnets of IPs banned from signup"""
+
+    ip = models.TextField()
+    """The IP address banned from signups"""
+
```
n-rook, line 285: IMO we should have a comment field here too, explaining why some IP is banned. Maybe an author field saying who banned them too if you can figure out a good way to set that

n-rook (changes requested):
Thanks for finishing this off

project/thscoreboard/users/other_views.py
```diff
@@ -18,7 +22,23 @@
 _USE_PASSCODE = True
 
 
+def check_ip_bans(methods_to_check: list):
+    def is_ip_banned(ip_to_check: str):
```
n-rook, line 26: nit: I don't think there's any reason for this to be a closure, so I would define it externally as a private function (that is, a function with a single underscore in front)

n-rook (approved):

project/thscoreboard/users/other_views.py
32th-System: I mainly put it there specifically so that the function is only available in the scope of the function that needs this functionality. It's more about scope for me

n-rook: that's not how I would do it but I don't have anything against it either as an idea, it's just a matter of taste. so that's fine

PR #170, opened 2022-12-27
Run uwsgi.reload() before running other commands

32th-System:
that way uwsgi already sees the code changes when the other commands are being run

### DIFF
--- a/project/thscoreboard/thscoreboard/autodeploy.py
+++ b/project/thscoreboard/thscoreboard/autodeploy.py
@@ -29,6 +29,7 @@ def do_deploy():
     # Something weird might happen if we try to do this concurrently.
     with _DEPLOY_LOCK:
         git_pull()
+        uwsgi.reload()
 
         management.call_command(
             'migrate',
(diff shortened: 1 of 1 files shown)


n-rook:
Unfortunately, I don't think this will work. uWSGI.reload() [gracefully reloads the site](https://uwsgi-docs.readthedocs.io/en/latest/Management.html#reloading-the-server). In other words, it waits for all current requests to complete, and then reloads the site.

However… This is a request! uWSGI isn't going to do anything until this function returns, no matter where you call reload(). As such, I don't think this change does anything.

32th-System:
Actually, it would be useful in one way. Because the request to reload is sent before any of the other commands get the chance to throw an exception, the request goes through even if some part of deployment fails. As such, I could simply try deploying again and it would see the code changes

n-rook:
Hmm. That's true… but is it necessarily a good idea? Maybe it is, but I'm not totally sure. If you can think of a scenario where you'd be happier for it to work this way, write it in a comment (explaining why we're calling reload() so early) and I'll approve the PR

PR #175, opened 2022-12-31
th12

raviddog:


project/thscoreboard/replays/game_fields.py
```diff
@@ -156,7 +179,20 @@ def GetFormatPower(game_id: str, power: Optional[int]) -> str:
     'th08': None,
     'th09': None,
     'th10': None,
-    'th11': 5
+    'th11': 5,
+    'th12': 4
+})
+
+_bomb_pieces = immutabledict({
```
n-rook, line 186: Add a comment saying what this dict is

project/thscoreboard/replays/game_fields.py
```diff
@@ -156,7 +179,20 @@ def GetFormatPower(game_id: str, power: Optional[int]) -> str:
...
+    'th01': None,
+    'th05': None,
```
n-rook, line 188: We support th02-04 too, so we should specify them as well. (Or, alternatively, I think I'd be OK with only including the games with bomb pieces in them here, and calling `_bomb_pieces.get(game_id, None)` instead.)

project/thscoreboard/replays/replay_parsing.py
```diff
@@ -530,6 +532,76 @@ def _Parse11(rep_raw):
...
+    rep_stages = []
+    for stage in replay.stages:
+        s = ReplayStage()
+        s.stage = stage.stage
+        s.score = stage.score * 10
+        s.power = stage.power
+        s.piv = (math.trunc(stage.piv / 1000)) * 10
+        s.lives = stage.lives
+        s.life_pieces = stage.life_pieces
+        #   fix zun fuckery
```
n-rook, line 554: lol

project/thscoreboard/replays/replay_parsing.py
```diff
@@ -530,6 +532,76 @@ def _Parse11(rep_raw):
...
+            stage.piv = None
+            stage.lives = None
+            stage.life_pieces = None
+            stage.bombs = None
+            stage.bomb_pieces = None
+            stage.graze = None
+
+    r_type = game_ids.ReplayTypes.REGULAR
+    if len(rep_stages) == 1 and replay.header.difficulty != 4:
+        r_type = game_ids.ReplayTypes.STAGE_PRACTICE
```
n-rook, line 588: I think if we're going to detect stage practice, we need a test that's stage practice, just to check.

n-rook (changes requested):
Looks basically good, just a few minor comments

project/thscoreboard/replays/game_fields.py
raviddog: games that don't include stage splits shouldn't ever be called in here, though theres no harm in it so i guess ill put em all in

project/thscoreboard/replays/replay_parsing.py
raviddog: theres already one, test_replay_parsing.py line 195

raviddog:
wait why do these tests fail it works on my local branch

raviddog:
the more i fiddle with this the more it breaks

project/thscoreboard/replays/replay_parsing.py
n-rook: I mean a test that actually tests this code in particular; that is, a test for UFO stage practice

project/thscoreboard/replays/test_replay_parsing.py
```diff
@@ -19,7 +20,7 @@ def testHard1cc(self):
         self.assertEqual(r.shot, 'ReimuA')
         self.assertEqual(r.score, 92245410)
         self.assertEqual(r.name, 'nrook   ')
-        self.assertEqual(r.replay_type, 1)
+        self.assertEqual(r.replay_type, game_ids.ReplayTypes.REGULAR)
```
n-rook, line 23: this is way easier to read btw, thanks for doing this

n-rook (changes requested):

n-rook:
python datetimes can be really crazy and even half of professional developers don't know how to handle them right. basically the best thing to do is to make everything an "aware datetime" by passing in a timezone immediately, because otherwise some library operations will assume the datetime represents local time. this tends to manifest as tests failing on github but passing locally (or the other way around) because the test env on github is UTC but your local pc probably isn't. fortunately it seems to be fine now

n-rook (approved):

PR #177, opened 2023-01-02
Convert timezones to UTC before serializing them.

n-rook:
This fixes an odd issue I had when working on the "reanalyze replay" function. Specifically, the way it was before, Django would only convert a naive datetime to an aware datetime when the instance was actually saved, which lead to weird issues when comparing the "old" replay row and the "updated" one. This also made Django log annoying warnings.

I am still not actually sure how Touhou itself handles timezones, though. Are the timestamps saved in replay files UTC, or are they local time?

### DIFF
--- a/project/thscoreboard/replays/create_replay.py
+++ b/project/thscoreboard/replays/create_replay.py
@@ -6,7 +6,6 @@ not with web entities like views or forms.
 
 from typing import Optional
 from django.db import transaction
-from django.utils import timezone
 
 from replays import models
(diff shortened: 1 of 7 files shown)


32th-System:
th07 does use localtime. Here's a small function decompiled from th07.exe
```c
void time_get_str(char *out)
{
  time_t utc;

  time(&utc);
  strftime(out, 6u, "%m/%d", localtime(&utc));
}
```

Issue #105, opened 2022-11-14
should we add a function to re-parse all replays


raviddog:

n-rook:
this is a good idea imo. since we haven't released yet we've been fine just adding new fields and saying "reupload if you want this data", but after we release we will want a way to add new fields according to replays we already have

PR #182, opened 2023-01-08
Allow reanalysis to actually update a replay on the site.

n-rook:
This change adds a button to the /reanalyze hidden route that updates the replay in question.

This way, if we change how derived fields are set on a replay, we can use this to update old replays. This is still not really what we want for #105 (what we want is to reparse all replays), but it's close; now that we have this it should be easy to finish that one off.


raviddog (approved):
Seems fine, though I'm not used to reviewing stuff so I'll trust that these tests handle things correctly

n-rook:
I'll write up a description of how I generally do code reviews at some point to give you more ideas

PR #183, opened 2023-01-08
read database name from environment variable if possible

32th-System:
To remove local changes from the staging website on the server


n-rook (approved):
yay

Issue #184, opened 2023-01-08
Remove private replays


n-rook:
Private replays are kind of pointless, and an extra category just makes the site more confusing. (After all, why would someone want to use the site if they don't want to share replays with other people?)

raviddog:
I can think of reasons, like using it for backup purposes or if they just don't want to be involved with the community. How much more complicated does keeping this feature make things?

n-rook:
My main concern with it was that it added complexity to the web design. We already have three different categories of replays: regular, tool-assisted and unusual (aka "anything else"); adding a fourth kind is something else users have to keep in mind. I know I've heard from people who are intimidated by how many options Lunarcast has. And I'm just not convinced it would really be useful to people.

If we did support private replays, I actually think the current implementation isn't good anyway. "private" isn't really a category in the way regular, TAS etc., are; it's really a different dimension altogether.

raviddog:
makes sense, so if we ever wanted to add it in the future it would need a separate framework rather than the implementation we currently have, gotcha

PR #185, opened 2023-01-08
Do not allow uploading private fields.

n-rook:
(Also, changes the name of "TAS" to "Tool-Assisted" replays.) #184


project/thscoreboard/replays/models.py
```diff
@@ -100,7 +100,11 @@ class Category(models.IntegerChoices):
     """
 
     PRIVATE = 4, pgettext_lazy('Category', 'Private')
-    """A private replay that isn't shown to anyone."""
+    """A private replay that isn't shown to anyone.
+
+    Private is deprecated; it will be removed soon. (If someone didn't
```
raviddog, line 105: We're in full "private" development mode without any users or external tools relying on our code, why do we need to depreciate stuff before taking it out?

raviddog (commented):
See comment on issue

project/thscoreboard/replays/models.py
n-rook: We have beta users now; what if one of them adds a private replay after I submit this PR but before it is released to production? This is unlikely, but it's not totally impossible, and it's best to get in the habit of being safe early (IMO)

n-rook:
yeah I won't merge this until we decide whether we really want to do it

raviddog (approved):

32th-System:
so, can we merge this now? The changes are approved

n-rook:
oh yeah we probably should, I've just been lazy. I'll pull in new changes and make sure it still works first, but after that I'll submit it

PR #186, opened 2023-01-11
add slowdown to display+parsing

raviddog:
how do i test a float


raviddog:
fixes #174

project/thscoreboard/replays/models.py
```diff
@@ -254,6 +254,9 @@ def GetDifficultyUrlCode(self):
     but technically there are 12 bytes allocated, so space is reserved just in case
     """
 
+    slowdown = models.FloatField(blank=True, null=True)
+    """Slowdown percentage in the replay"""
```
n-rook, line 258: Specify the range in the comment. (In other words does it go from 0 to 1, or from 0 to 100?)

n-rook (changes requested):
Add at least one test, please!

n-rook (approved):

PR #188, opened 2023-01-11
add upload replay button to header

### LINKED ISSUE #172: put the upload button on the header

raviddog:
fixes #172


n-rook (approved):
This seems fine, but please add a screenshot when adding things like this! Otherwise folks can't really know how it will look before approving it, unless we actually download the branch ourselves and run it locally.

PR #199, opened 2023-01-28
Th13

raviddog:
TestTableFields is causing issues

td has varying life piece threshholds, which means the existing "display as current/max" method doesnt work
so i need to bring back the life pieces column, but only for games with varying life pieces
which means games that do have the life piece field might need life pieces visible to be none
and the test doesnt like that


project/thscoreboard/replays/models.py
```diff
@@ -419,6 +419,19 @@ class Meta:
     th09_p2_score = models.IntegerField(blank=True, null=True)
     """The score stored for player 2 in the stage split in TH09"""
 
+    th13_trance = models.IntegerField(blank=True, null=True)
+    """Trance gauge value at the stage split in TH13
+    This value ranges from 0 to 600, with 1 ingame trance level equalling 200 points"""
```
n-rook, line 424: Nitpick: In Python, when you have a multiple-line docstring like this, it is conventional to put the last `"""` on a new line by itself. For example:

```
"""Trance gauge value at the stage split in TH13

This value ranges from 0 to 600, with 1 ingame trance level equalling 200 points
"""
```

project/thscoreboard/replays/game_fields.py
```diff
@@ -5,6 +5,11 @@
 from typing import Optional
 from . import game_ids
 
+#   These table fields configure the display of the stage split columns
+#   These may differ from whether the replay file actually has the field or not
+#       depending on whether it is relevant to show that field
+#   eg. life_pieces will almost always be False as displaying it is integrated into Lives
```
n-rook, line 11: (I'm leaving this comment here because there's no way to leave a general comment as part of a code review...)

As you know, this PR breaks the test in test_view_replay that checks that GetGameField corresponds with the fields on an actual replay, because it doesn't show life_pieces in some cases where it's shown on the website. This test exists to make sure we don't accidentally forget to display something, or display something that isn't set; but we are intentionally not showing life_pieces sometimes even though it's set, so we are OK with this and don't want the test to fail. You don't need to overthink this one: this case is an exception to the normal state of things, so you should just edit test_view_replay and make it be OK if life_pieces is set on a replay, even if GetGameFields() doesn't have life_pieces set to true. (And add a note in the test explaining why you put that exception in.)

n-rook (changes requested):
This looks basically good, only real issue is the failing test

n-rook (approved):

PR #201, opened 2023-02-06
Allow the site to respond to traffic from local.silentselene.net.

n-rook:
This address is just 127.0.0.1, but I think it can be more convenient to access this way.


32th-System:
so would "local.silentselene.net" be set in /etc/hosts?

n-rook:
Nah, I literally went and put an A record at it for 127.0.0.1.

32th-System:
then you should also add an AAAA record for ::1

n-rook:
Done

PR #209, opened 2023-02-08
Add systemd timers for deleting old data and backups

32th-System:


ref/server/systemd/silentselene-delete-backups.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=silentselene: clear out backups older than a week
```
n-rook, line 2: Could we do a month? A week makes me a little nervous, it's easy for me to imagine a situation where we want to restore information from more than a week ago

ref/server/systemd/silentselene-delete-old-data.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
```
n-rook, line 1: We should do this on staging too

n-rook (approved):

ref/server/systemd/silentselene-delete-old-data.service
32th-System: Already on the server. Should I put the staging files on here too? Because I didn't upload silentselene-staging.service either but I did upload silentselene.service

n-rook: hmm... I think so? I don't see why they shouldn't be here. I'll leave it up to you though. (And it'd be fine to do that in another PR if you want)

PR #217, opened 2023-02-09
Allow signups with early access passcodes disabled

32th-System:


n-rook (approved):
Looks good. Before submitting, make sure to check that you can't sign up with the wrong passcode. Looking at the code I think this should still work properly, but I wasn't 100% sure.

Issue #203, opened 2023-02-06
Allow users to delete their accounts


n-rook:
It is currently not possible for users to delete their own accounts.

We need to be careful about how to implement this, but I think it is clearly something we should implement, both for legal and ethical reasons.

32th-System:
Regarding the implementation, I have these questions that I think should all be answered for the implementation. There might be more, but these are the ones I can think of
- How should old replays from the deleted user be affected?
- What should happen with the username? How long until someone else could use that name?
- What if the same person wants to sign up again? We can verify if it's the same person who wants to sign up again based on the e-mail address. What if the username of a deleted account is up for grabs again but the person who deleted again wants to sign up again with that name? How should that affect the other person who took the name?
- What about regular bans? In what ways should the answers to the above questions change if it's not a regular account deletion but a regular ban?

n-rook:
Those are good questions. I'll lay out my design here.

Deleting an account should delete all information related to the account. This includes all of the user's replays. It's unfortunate to lose information like that, but I don't think it's right to keep that data if the person doesn't want us to have it anymore. Similarly, the username will become available if a new user wants it. (The email address will be available too, in case the user decides to come back.)

We should not immediately delete an account on request. After all, the user might change their mind a few days later. Or, the account might have been hacked, and the person behind it might not actually want it to be deleted. Instead, we should mark the account as "to be deleted". While the account is "to be deleted", the user can still log in, but the only thing they should be able to do is tell us they changed their mind and don't want the account deleted anymore.

If an account is in the "to be deleted" state, of course the username can't be reused, but we should make it similar to the account really being deleted in all other respects; notably, we should hide the replays so it looks like they were deleted.

If someone deletes their account, and someone else takes the username, and then the first person wants to come back with the username, too bad: it now belongs to the second person. I think this is the only sensible way to handle things. (This is why I want a "grace period" before actually deleting the account, so if the user has a change of heart they can keep their account.)

For bans, I think it's important to let the banned user delete their account, but it's just as important to not let them use this to get around a ban. So, I think if a banned user tries to delete their account, we should delete the user information, replays and so on, but keep their username and email address in a new "banned users" table, so they can't be reused. (I do think we are fine here from a GDPR standpoint.)

32th-System:
question about the "to be deleted" state: do the replays stay visible or should they be put in a state similair to private replays?

n-rook:
[...]

PR #219, opened 2023-02-09
Have cleanup_old_data delete user accounts after 60 days pass.

n-rook:
#203 

This PR can only be merged after #218. Its merge target is set as #218 to make the diff look OK.


project/thscoreboard/users/models.py
```diff
@@ -20,6 +21,9 @@
 class User(auth_models.AbstractUser):
     """A user."""
 
+    _DELETED_ACCOUNT_TTL = datetime.timedelta(days=60)
```
32th-System, line 24: would make the official length of time for which deleted user accounts are kept 90 days (accounting for database backups)

project/thscoreboard/users/models.py
```diff
@@ -54,6 +58,23 @@ class Meta(auth_models.AbstractUser.Meta):
...
+        earliest_surviving_time = now - User._DELETED_ACCOUNT_TTL
+        logging.info('Deleting accounts marked for deletion before %s', earliest_surviving_time)
+
+        count = 0
+        for user in User.objects.filter(
+            is_active=False,
+            deleted_on__lte=earliest_surviving_time
+        ):
+            with transaction.atomic():
+                user.delete()
```
32th-System, line 74: Just to be sure: is this really guaranteed to work? Because when I tried deleting user objects from pgAdmin4 or the Django admin panel it would complain about replay or visit objects still referencing the user object

project/thscoreboard/users/models.py
n-rook: Yep. We should actually claim something longer (we will probably back up to remote storage at some point, which will probably add another 30 days) but this'll be the real number. It seems reasonable enough.

project/thscoreboard/users/models.py
n-rook: I tested it and I think it should work, although I think I forgot to commit the migration until right now... Replays should be good but visits were definitely broken, that's why I changed them to ON_CASCADE. I added a big test for this just now, since you mentioned it (and it would definitely be super easy to break).

n-rook:
This is ready to merge now

PR #221, opened 2023-02-10
have the sign up page make the user agree to the terms of use and privacy policy

32th-System:


n-rook:
I don't think we need to do this. The terms of use and privacy policy are freely accessible on the website, after all. Most of the legal basis for what we do is based on legitimate interest, or is based on consent where the consent is intrinsic in performing the action (for example, "if you upload a replay, we will store the replay you uploaded"). And I don't really want to do it, because it adds friction and makes it more annoying for people to sign up.

PR #222, opened 2023-02-10
Update privacy policy

32th-System:


project/thscoreboard/replays/templates/replays/docs/privacy_policy.html
```diff
@@ -96,11 +96,22 @@ <h3>Public Information</h3>
 
     <h2>How Long We Keep Information</h2>
     <p>
-        We generally discard information about you when it's no longer needed for the purposes for which we collect and use it. For example, if content is deleted, we stop storing it in our database within 30 days after its deletion. After this time period is up, we may still store it in caches or backups, but these backups will eventually be purged.
+    We generally discard information about you when it's no longer needed for the purposes for which we collect and use it, though all deleted data is kept for up to 60 days in backups. Replays deleted by the user are not kept in the database for any longer than that.
     </p>
     <p>
-        We may store IP addresses and other information for longer periods of time if we deem it necessary to protect our site.
+    User visits are recorded in the database for users signed in with an account, as well as users browsing the site anonymously. If a visit is recorded for a user, who is signed in with their account, the visit will be associated with said user account. Information about visits is kept for 41 days.
```
n-rook, line 102: I think we should refrain from saying "user visits", since people don't know what that means--- what they do know is IP addresses. I would say something like this:

We store the IP addresses of visitors to the site, and cross-reference them with user accounts. This information is deleted after 120 days.

project/thscoreboard/replays/templates/replays/docs/privacy_policy.html
```diff
@@ -96,11 +96,22 @@ <h3>Public Information</h3>
...
     </p>
+    <p>
+    Should a user choose to delete their account, then all information about the deleted account, including any replays still associated with that account, is kept for an additional 60 day period, to allow the user to undo their account deletion. Visits are not affected by this
```
n-rook, line 105: Drop "Visits are not affected by this", since we cover that above.

project/thscoreboard/replays/templates/replays/docs/privacy_policy.html
```diff
@@ -96,11 +96,22 @@ <h3>Public Information</h3>
...
+    </p>
+
+    <h3>Suspended user accounts</h3>
+    <p>
+    Should a suspended user account be deleted, then all information about the account, except for the e-mail address, username and reason for suspension is considered relevant for the suspension and will be kept for at least the duration of the suspension, even if it exceeds 120 days. Any other information will be deleted after the 120 day period. This is to prevent suspended users from being able to evade their suspension with the same e-mail address.
```
n-rook, line 110: I think you wrote the opposite: It's not that all information is kept, it's that all information is deleted. I would revise the first sentence as follows:

A suspended user account can still be deleted. However, if this happens, the username, email address, and reason for the suspension are kept for the duration of the suspension, even if it exceeds 120 days.

n-rook (commented):
Looks mostly good, just a few comments. Thanks for doing this

n-rook (approved):

PR #230, opened 2023-02-11
Revert "Add a "request invite" page and hide the signup page from the site.

n-rook:


n-rook:
oops: we should keep the passcode in staging

PR #231, opened 2023-02-11
Control passcode on/off based on an environment variable.

n-rook:


project/thscoreboard/thscoreboard/settings.py
```diff
@@ -38,6 +51,9 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = not IS_PROD
 
+# Whether or not to require a passcode to sign up for the site.
+REQUIRE_PASSCODE = GetBooleanFromEnvironment('REQUIRE_PASSCODE', True)
```
32th-System, line 55: this can also be done with eval()

n-rook: Yeah, but eval is a bit scary, I prefer to avoid it if possible.

PR #247, opened 2023-02-12
Extract fetching shot, game and route to a saner place.

### LINKED ISSUE #244: Trying to reanalyze a regular th08 replay fails
It tries to assign the route string "Final A"/"Final B" to the route field and fails
Only happens on full game runs

n-rook:
The previous way this was done was kind of ad hoc and confusing, which led to bugs. I'm still not totally happy with this approach either but it should at least be better.

Fixes #244


raviddog (approved):
Nice

PR #248, opened 2023-02-12
All encodings in userdata should be specified as SJIS

32th-System:


n-rook:
This makes sense to me, but could you add a test? It'd be great if we could find a replay where the name is only parsed properly with shift-JIS

32th-System:
added it. Any replay that was played without the NDT static English patch would have this issue

n-rook (approved):

PR #251, opened 2023-02-13
Th14

raviddog:
let me know if you want the game checking code not to have funny japanese character in it and instead replace it with the actual byte number, though i think i've left ample documentation


n-rook (approved):
Great!

n-rook:
And no, I think it's fine to have Japanese characters in the codebase; it makes it clearer what's actually being expected.

PR #252, opened 2023-02-13
Don't include download link for replays without/with bad replay files on replay tables

32th-System:


.github/workflows/deploy_staging.yml
```diff
@@ -7,7 +7,9 @@ on:
 
 jobs:
   request_deployment:
-
+    
+    if: ${{ secrets.STAGING_DEPLOYER_PASSWORD }}
```
n-rook, line 11: This seems unrelated, send out a separate PR

project/thscoreboard/replays/templates/replays/replay_table.html
```diff
@@ -50,8 +50,12 @@
             <td>
                 {{replay.created | date:"Y-m-d"}}
             </td>
-            <td>
+            <td style="text-align:center;">
+            {% if replay.is_good %}
```
n-rook, line 54: This makes sense to me, good idea!

n-rook (approved):

.github/workflows/deploy_staging.yml
n-rook: Wait you already did lol

32th-System: wait did I forget that I did this PR and the other one on the same main branch on my fork. lol

PR #255, opened 2023-02-14
fix #140 for real

### LINKED ISSUE #140: Don't try to deploy to staging on a fork of this repository
Apparently the cbuild is trying to deploy to staging even on forks (and then it doesn't work because it doesn't have the staging credentials)

32th-System:


32th-System:
can't check if a secret is present using that if statement.

n-rook (approved):
whoops!

PR #256, opened 2023-02-15
make all replay stages 0-indexed

### LINKED ISSUE #181: mof replays are 1-indexed
I added

```
    def testStages(self):
        r = ParseTestReplay('th10_normal')
        self.assertEqual(r.stages[0].stage, 0)
```

to test_replay_parsing.py and it failed.

They should be 0-indexed, since the field is documented as 0-indexed (and some other code will probably expect it)

jgdhs27:
Make all replay stages 0-index, in order to be consistent. Fixes #181

### DIFF
--- a/project/thscoreboard/replays/kaitai_parsers/th10.py
+++ b/project/thscoreboard/replays/kaitai_parsers/th10.py
@@ -50,7 +50,7 @@ class Th10(KaitaiStruct):
             self._read()
 
         def _read(self):
-            self.stage = self._io.read_u2le()
+            self.stage_num = self._io.read_u2le()
             self.unknown_1 = self._io.read_u2le()
             self.unknown_2 = self._io.read_u4le()
             self.next_stage_offset = self._io.read_u4le()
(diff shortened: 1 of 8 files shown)


32th-System:
you forgot to change test_reanalyze_replay.py

jgdhs27:
fixed. tests are green.
but now that ive implemented this, i actually think 1-indexed is better...
`stage2 = models.ReplayStage.objects.get(replay=replay, stage=1)` is super confusing

n-rook:
Looks good.

For 0-index vs 1-index, I don't really have strong feelings about it. The only thing I cared about was that it was consistent. If you feel like changing it you could talk to raviddog and see what they think.

I think it's good that ReplayInfo returns a list (versus a dict) because it retains the concept of an order of stages. Well, not that it matters that much the exact formatting of the ReplayInfo, since it is used primarily as an intermediate structure to handle information about a replay that hasn't been saved to the database yet.

I do think there's an argument that something other than an integer should be stored in the database, but every other option has its own downsides too. An integer is ordered, which is nice.

n-rook (approved):

PR #258, opened 2023-02-16
make all replay stages 1-indexed

jgdhs27:
lol

also, changed `GetFormatStage` so that extra stages are formatted as "Extra" instead of "7"


project/thscoreboard/replays/game_fields.py
```diff
@@ -353,24 +353,27 @@ def GetGameLifePieces(gameid: str):
 
 def GetFormatStage(game_id: str, stage: Optional[int]) -> str:
     if stage is None:
-        return ""
+        return ''
+    format_ex = 'Extra'
```
n-rook, line 357: Since this is a constant, we should declare it at the top-level in all caps, like `_EXTRA_NAME`. If it is declared as a local variable like this, I expect it to change, so I get surprised when it doesn't change.

n-rook, line 357: Could you upload a screenshot of what this looks like in practice? I'm worried "Extra" will take up too much room on the table compared to just numbers, so maybe "Ex" would be better. But I'm not sure, so I'd like to see it.

project/thscoreboard/replays/replay_parsing.py
```diff
@@ -80,7 +80,7 @@ class ReplayInfo:
     replay_type: int
     route: Optional[str] = None
     spell_card_id: Optional[int] = None
-    stages: list[ReplayStage] = dataclasses.field(default_factory=list)
+    stages: List[ReplayStage] = dataclasses.field(default_factory=list)
```
n-rook, line 83: It's better to use the lowercase ones; see https://peps.python.org/pep-0585/.

n-rook (changes requested):
It looks like you need to update your branch to include the latest changes from the main branch, `nrook:main`. iirc you can do this by checking out your branch, running `git fetch` to fetch the latest remote changes, and then running `git merge nrook/main`

project/thscoreboard/replays/game_fields.py
raviddog: would it suffice to declare it right above the function? since it's only being used here I wouldn't want it to be too far away

raviddog: While we're doing this it would be nice to have Phantasm as well, I guess it could be "Ph" and then we could use "Ex"
I think add an optional difficulty argument to the function to achieve this

raviddog:
fixes #181

project/thscoreboard/replays/game_fields.py
n-rook: Yes, that would be fine

jgdhs27: wasnt too sure since the constant is only use right here. ill follow raviddog's suggestion

jgdhs27: > While we're doing this it would be nice to have Phantasm as well, I guess it could be "Ph" and then we could use "Ex" I think add an optional difficulty argument to the function to achieve this

I completely forgot phantasm existed. added.

project/thscoreboard/replays/replay_parsing.py
jgdhs27: oh cool didnt know that. ive always used `List` but im on an older python, looks like its changed.

jgdhs27:
merging with the "0-index" commit reverted a lot of my changes. fixed it by hand, hopefully didnt miss anything, tests are green at least

project/thscoreboard/replays/game_fields.py
jgdhs27: image
looks good to me

jgdhs27:
In order to display Ph properly, I needed to give it a distinct stage number. Previously, ex and ph both had stage number 7. I changed it so that ex has 7 and ph has 8. Feels like a bugfix.

raviddog (approved):

n-rook (approved):

PR #261, opened 2023-02-18
Parse th15-th18

jgdhs27:


project/thscoreboard/replays/kaitai_parsers/th06.py
```diff
@@ -15,7 +15,7 @@ def __init__(self, _io, _parent=None, _root=None):
         self._read()
 
     def _read(self):
-        self.header = Th06.Header(self._io, self, self._root)
+        self.file_header = Th06.FileHeader(self._io, self, self._root)
```
n-rook, line 18: This seems like it doesn't belong in this PR

project/thscoreboard/replays/replay_parsing.py
```diff
@@ -795,6 +687,213 @@ def _Parse14(rep_raw):
...
+            power=current_stage_end_data.power,
+            piv=(math.trunc(current_stage_end_data.piv / 1000)) * 10,
+            lives=current_stage_end_data.lives,
+            life_pieces=current_stage_end_data.life_pieces,
+            bombs=current_stage_end_data.bombs,
+            bomb_pieces=current_stage_end_data.bomb_pieces,
+            graze=current_stage_end_data.graze,
+        )
+        if next_stage is not None:
+            # The end-of-stage data does not add the stage's clear bonus.
```
n-rook, line 871: Huh, is this only the case in UM and not in other games in the same era?

project/thscoreboard/replays/replay_parsing.py
```diff
@@ -166,50 +161,33 @@ def _Parse07(rep_raw):
...
-            stage = rep_stages[i]
-            stage.power = None
-            stage.lives = None
-            stage.bombs = None
-            stage.point_items = None
-            stage.graze = None
-            stage.piv = None
-            stage.th07_cherry = None
-            stage.th07_cherrymax = None
+    for (i, current_stage), (j, next_stage) in zip(
```
n-rook, line 173: I do like this refactoring, it looks much cleaner this way to me.

project/thscoreboard/replays/replay_parsing.py
```diff
@@ -795,6 +687,213 @@ def _Parse14(rep_raw):
     return r
 
 
+def _Parse15(rep_raw) -> ReplayInfo:
```
n-rook, line 690: This is optional, but have you thought about trying to reuse some of the code between the different games? I know I've heard that the newer games all have similar structures in the replays (as opposed to the older games which do weird stuff), so it seems like maybe it would be possible. And it'd make it easier to add new data parsing features later.

n-rook (changes requested):
I think this PR is way too big to review! In my experience, the difficulty of actually giving a good review that can catch issues scales about O(n^2) with the size of the review diff. And the diff here is something like 800 lines (not counting the generated code which doesn't count because I don't need to review it). At the very least, refactoring the existing replay parsing code and adding new games needs to be in separate PRs.

Of course, the downside of smaller reviews is that it adds to the time to get a change in. I have used "stacked PRs" before to fix this, where my first PR targets the main branch, but the second PR targets the first branch. (And then I never merge the second PR before the first one, I just wait for the first one to be approved.) This can be confusing but it's up to you if you want to use it. (Well, really I don't think the different changes here depend on each other anyway.)

project/thscoreboard/replays/replay_parsing.py
32th-System: they are similair but the finer layout of the structs does change between every game and it's easier to just spell it out again. There is some reuse though with the th_modern struct

project/thscoreboard/replays/replay_parsing.py
raviddog: Agree, this is much nicer than my mess

project/thscoreboard/replays/replay_parsing.py
raviddog: It would be possible I think, and we can look into it later but for now each game has enough of its own oddities that we should focus on functionality first, getting everything up and running

project/thscoreboard/replays/game_fields.py
```diff
@@ -242,6 +330,10 @@
     'th12': _table_fields_th12,
     'th13': _table_fields_th13,
     'th14': _table_fields_th14,
+    'th15': _table_fields_th15,
+    'th16': _table_fields_th12,
```
raviddog, line 334: typo here :)

raviddog (changes requested):
Apart from the typo, everything looks good

Theres a couple things you can add but they're not necessary in this PR and we can do them later. If you choose to do them later, open an issue so we don't forget.

- th16, th17 and th18 have spell practice, checks for these aren't present
- there are some game specific fields that haven't been parsed
  - th16 season
  - th17 tokens and hyper bar
  - th18 cards
  - The latter two here need some additional groundwork done for their implementation, so I feel they can be left separately

Other than that, if everything else looks good on your end then we're good to go

project/thscoreboard/replays/replay_parsing.py
raviddog: UM is the first game to have end-of-stage data so this problem can't occur in previous games

(I originally made this comment then misread the code and deleted it)

project/thscoreboard/replays/replay_parsing.py
jgdhs27: I agree with 32th here, while there is a lot of overlap, each struct is subtly different.

project/thscoreboard/replays/kaitai_parsers/th06.py
jgdhs27: yeah, i can make a separate PR for that one

project/thscoreboard/replays/game_fields.py
jgdhs27: fixed

raviddog (approved):
Awesome

jgdhs27:
added a field for th16 season power

n-rook:
Yeah from what I can tell this looks good. Split the refactoring into a separate PR (from what I can tell this should be easy, since it's split into different commits anyway) and it'll be good to go

jgdhs27:
Rebased + added spell practice parsing

n-rook (approved):
Looks great, thanks for all this!

PR #262, opened 2023-02-18
Parse the replay only once when uploading

32th-System:
this also accidentally adds a full edit feature for all non PC-98 games. Follow-up PR adding the things required to make it official coming soon


n-rook:
I'm not sure this is a good idea, for two reasons.

The first reason is that it complicates the user experience. Previously, from the user's perspective, when they upload a replay, the replay isn't added to the site until they fill out the follow-up page. If they realize on parsing it that they uploaded the wrong replay, they can just leave; they don't have to do anything to cancel the upload or delete the pending file. After this PR, the idea of a pending replay is introduced to the user. If they upload the wrong replay, they need to actually take an affirmative action to get rid of it; otherwise, it will clog up their profile page and stick around indefinitely.

The second reason is that it complicates the "Replay" table. Previously, any row in the Replay table was a valid replay, or at least a replay with all required information filled out. Now, though, there will be lots of replays in the Replay table with bogus values for fields like "score" or "is_clear". This makes the database more confusing, because it limits what we can know about a Replay instance; we don't know that some of these fields have legitimate information in them or not. This makes it easier for us to introduce bugs later by getting confused.

Of course, there is one major benefit to this PR too, which is it makes the code simpler. The TemporaryReplayFile table is gone, and a hypothetical future "edit" page would look basically the same as the "finalize publish" page does now. These are both nice, but I don't think they outweigh the two factors above. (I don't think it really matters whether we parse a replay once or twice, since parsing replays is quite cheap anyway.)

I do want other people to weigh in on this too. if everyone thinks this is a good idea but me, we can still go ahead and do it.

32th-System:
I can add a TTL to pending replays I just don't know how that whole system works lmao

32th-System:
So here are my more nuanced thoughts:

Regarding the first issue: I could keep the pending replays on the user's profile page, but also have pending replays expire and get deleted after 30 days automatically. That way, while the concept of a pending replay still exists, the user doesn't have to worry about it. I could also add the TTL to pending replays and not show pending replays on the profile page, which would remove the concept from the user entirely

As for the second issue: I personally don't care. I pretty much never look through the replay tables on the production database. I rarely even look at the replay table in my local database. If I needed to look through my local database, I can deal with the pending replays myself, but on the production database I don't need to look through the replay table.

I'd still like to know the thoughts of other maintainers

n-rook:
Yeah, as you mention the first issue isn't really core to the PR; you could maintain the current interface while still making most of these changes.

The second issue isn't about personally looking at the database; it's about writing code against the database. The more complex the database, the higher the rate of bugs will be in code that is written against the database. And on a replay site, almost all code is written against the database. This PR puts information into the database that is actually wrong: fields like `is_clear` receive preliminary values that aren't really accurate. I think it would probably be better to set these fields to null instead, but even if we did that, now code that deals with replays has to worry about these fields possibly being null. The previous code did need an extra table, but that table compartmentalized that information, so that most code did not have to concern itself with replays that were in the process of being uploaded.

A big part of keeping a codebase healthy and easy to work with is ruthlessly compartmentalizing sources of complexity; that's really the source of my concern.

32th-System:
> fields like is_clear receive preliminary values that aren't really accurate. I think it would probably be better to set these fields to null instead, but even if we did that, now code that deals with replays has to worry about these fields possibly being null.

I mostly did that to get around constraints but I don't really know how they work. The proper way to do things would be to have constraints that force values like is_clear to be None while having the normal constraints if the replay isn't pending.

32th-System:
One more thing regarding the constraint thing though: if fields like is_clear can't be None if the replay isn't pending, then anything that doesn't want to deal with pending replays can simply filter for replays where the category is not "Pending" and then that code can still assume that is_clear can only ever be true or false.

Follow up thought: I made Pending a category rather than it's own True/False field because I wanted to make new use of the already existing code for private replays. Should "Pending" be it's own field in the replay table or bunched in with categories?

raviddog:
I think I agree with n-rook's second point. We don't want things clogging up the Replay table that don't serve the same functionality as the Replay table. Whether this change goes ahead or not I don't mind, but it should be integrated in a way where future code edits don't have to worry about accidentally tripping over some additional constraint when dealing with the database.

32th-System:
I have added constraints so that
- Any code that excludes pending replays when iterating over the replay table never has to deal with things that couldn't be null before suddenly being able to be null
- is_clear has to be null if the replay is pending and can't be null if it isn't
Is there anything else I could do to mitigate the second point?

32th-System:
Pushed the changes to add a TTL to pending replays. Also, my constraints from 2 commits ago don't prevent pending replays to have a value for is_clear, but they do still prevent is_clear from being null if the replay is not pending

n-rook:
Those changes definitely help, but my original concern still stands.

Another way to think about it is that after this change, the database is less normalized. I never studied database normalization academically, so my understanding of it is only conversational and intuitive, and I can't really talk about it formally well. But from this perspective, it's believed that it makes the database harder to deal with if it contains redundant data. In this case, there are two new forms of redundant data introduced. The first is that if the replay is pending, certain fields always have certain values. This is alleviated partly by the use of constraints, but it still adds to the complexity of the table. The second form of redundancy is that lots of fields on the Replay table are functions of the replay itself. Of course, this redundancy is also present in non-pending replays, but it is necessary there; we need these values in the database so we can make queries against them. But for pending replays, the only benefit to storing this information in the database is to prevent us from having to re-parse the replay. And parsing replays is so fast that this benefit is very small.

Like I said earlier, if you can convince both raviddog and morth that this change is a good idea, I'm happy to be overruled and let it in. But I really do think it is cleaner to keep pending replays in a separate table, so that most of the codebase never has to worry about their existence.

32th-System:
I never heard of database normalization until today. I'm not really sure how to articulate this, but in essence, the thought of the second point never occured to me when I made this PR, and I still don't think it's a big problem, especially with constraints and tests in place. Like sure, bugs might be more likely to occur with this change because it makes the main replay table more complicated, but bugs can also happen without this change. I think that the difference is very very small though, especially with proper constraints and test in place and I think the benefits do outweigh the downsides.

As for what the actual performance benefits are, I want to run the tests for that on the production server, since it has much worse specs than any of our dev machines.

32th-System:
I decided to run tests on the server. The performance difference is very small, only about 0.05 seconds even on a low end server, and I could easily implement an editing feature in a different way. I'm ready to close this now since this wouldn't be worth it, and also the commit history on this PR is a mess now

32th-System:
I am closing this now and I am saving the important commits in case we decide that we want to unearth this again. If we wanted to merge this, I'd want to remake the PR anyways.

PR #264, opened 2023-02-19
Refactor: Make use of next stage explicit

jgdhs27:
Previously, stage parsing was done in two steps: First create objects containing start-of-stage data and then change them in place to contain end-of-stage data. I reworked this so that it's a 1-stage process.

For th10-th14, I use `zip(stages, stages[1:])` to iterate through pairs of stages. th06-th08 have dummy stages, so I have to first enumerate, then filter dummies, then zip. th09 is even more complicated so I decided not to change it, and added a comment instead.

Tests are unchanged, tests still pass.

### DIFF
--- a/project/thscoreboard/replays/replay_parsing.py
+++ b/project/thscoreboard/replays/replay_parsing.py
@@ -106,34 +106,26 @@ def _Parse06(rep_raw):
     shots = ["ReimuA", "ReimuB", "MarisaA", "MarisaB"]
 
     rep_stages = []
-    for i, score in enumerate(replay.stages):
-        if replay.header.stage_offsets[i] != 0:
-            s = ReplayStage()
-            s.stage = i + 1
-            s.score = score.score
-            s.power = score.power
-            s.lives = score.lives
-            s.bombs = score.bombs
-            s.th06_rank = score.rank
-            rep_stages.append(s)
 
-    # adjust stage data to be end-of-stage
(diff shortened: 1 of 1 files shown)


project/thscoreboard/replays/replay_parsing.py
```diff
@@ -106,34 +106,25 @@ def _Parse06(rep_raw):
...
-        else:
-            stage = rep_stages[i]
-            stage.power = None
-            stage.lives = None
-            stage.bombs = None
-            stage.th06_rank = None
+
+    enumerated_stages = list(enumerate(replay.stages))
+    enumerated_non_dummy_stages = list(filter(
+        lambda e: replay.header.stage_offsets[e[0]] != 0, enumerated_stages
```
n-rook, line 112: I think this would be more readable as:
```
enumerated_non_dummy_stages = [
    e for i, e in enumerate(replay.stages) if replay.header.stage_offsets[i] != 0
]
```

I think it's better to use list comprehensions over the functional programming functions when possible; it just looks cleaner and more like regular Python style.

n-rook (changes requested):
Looks great, just one minor comment

project/thscoreboard/replays/replay_parsing.py
jgdhs27: nice suggestion

n-rook (approved):

PR #267, opened 2023-02-19
Show spell card names on spell practice replays

32th-System:


project/thscoreboard/replays/spell_names.py
```diff
@@ -0,0 +1,967 @@
+from immutabledict import immutabledict
+from django.utils.translation import get_language
+
+spell_names_en = immutabledict({
+    'th08': tuple([
```
n-rook, line 5: Instead of `tuple([a, b, c])` you can just do `(a, b, c)`

project/thscoreboard/replays/spell_names.py
```diff
@@ -0,0 +1,967 @@
...
+        '六鼓「オルタネイトスティッキング」',
+        '七鼓「高速和太鼓ロケット」',
+        '八鼓「雷神の怒り」',
+        '「ブルーレディショー」',
+        '「プリスティンビート」',
+    ]),
+})
+
+
+def get(game_id: str, spell_id: int):
```
n-rook, line 960: Add a docstring

project/thscoreboard/replays/views/view_replay.py
```diff
@@ -49,6 +50,7 @@ def replay_details(request, game_id: str, replay_id: int):
         'shot_name': replay_instance.shot.GetName(),
         'difficulty_name': replay_instance.GetDifficultyName(),
         'game_id': game_id,
+        'spell_name': spell_names.get(game_id, replay_instance.spell_card_id),
```
n-rook, line 53: opt: consider defining a GetSpellCardName() function on the replay model itself instead, I feel like that could be convenient to have around

n-rook (changes requested):
Looks great; only minor comments

(Also I freaked out because I saw a 1000 line diff before I realized why it was so long lol)

project/thscoreboard/replays/spell_names.py
32th-System:
```
C:\Users\thc>python
Python 3.11.2 (tags/v3.11.2:878ead1, Feb  7 2023, 16:38:35) [MSC v.1934 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> a = tuple("a", "b", "c")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: tuple expected at most 1 argument, got 3
```

32th-System: oh you meant without the tuple keyword

n-rook (approved):

PR #270, opened 2023-02-20
merge getformatlifepieces into getformatlives

raviddog:


project/thscoreboard/replays/game_fields.py
```diff
@@ -423,16 +395,34 @@ def GetFormatPower(game_id: str, power: Optional[int]) -> str:
 })
 
 
-def GetFormatLives(game_id: str, lives: Optional[int], life_pieces: Optional[int]) -> str:
+def GetFormatLives(game_id: str, lives: Optional[int], life_pieces: Optional[int], extends: Optional[int] = 0) -> str:
```
n-rook, line 398: `-> Optional[str]`, right?

project/thscoreboard/replays/game_fields.py
```diff
@@ -423,16 +395,34 @@ def GetFormatPower(game_id: str, power: Optional[int]) -> str:
...
+    lives_str = ""
+    life_pieces_str = ""
```
n-rook, line 400: You actually shouldn't need to declare these two variables up here like this. They're set in every code path anyway.

project/thscoreboard/replays/game_fields.py
```diff
@@ -423,16 +395,34 @@ def GetFormatPower(game_id: str, power: Optional[int]) -> str:
...
+
     if lives is None:
-        return ""
-    total_life_pieces = _life_pieces[game_id]
-    if total_life_pieces is None:
-        return str(lives)
+        return None
+    else:
```
n-rook, line 404: Since you `return None` if lives is None, you don't need to put anything in an else clause. After all the whole rest of the function will only execute if lives is not None.

project/thscoreboard/replays/game_fields.py
n-rook, line 398: Docstring plz

n-rook (changes requested):
Looks good, just a few minor comments

project/thscoreboard/replays/game_fields.py
raviddog: I like having them here as kind of a clear "these are the strings I will be using" sort of thing

n-rook (approved):

PR #271, opened 2023-02-20
th16 season

raviddog:
adds th13 trance to the frontend too and fixes table typos


project/thscoreboard/replays/models.py
```diff
@@ -462,6 +462,9 @@ class Meta:
     opted not to specify a game for its name
     """
 
+    th16_season_power = models.IntegerField(blank=True, null=True)
+    """Value of the season gauge in TH16"""
```
n-rook, line 466: It'd be good to list the min and max values of this, I think

n-rook (approved):

project/thscoreboard/replays/models.py
raviddog: I'm not entirely sure of the values here, from what I remember it ranged from 0 to 1200, but then I also remember something about different season levels having a different amount required or something. In the end I'm not confident enough to make an exact statement about these.

PR #274, opened 2023-02-21
Proposal: Use black formatter

jgdhs27:
[Black](https://pypi.org/project/black/) is a code formatter that runs every time you save a file. If you use black, you don't have to think about formatting at all. It's really convenient.


32th-System:
could you reset the files in replays/kaitai_parsers, replays/migrations and users/migrations and tell it to ignore the files in those folders? Those files are all autogenerated. Still, I'm unsure about this

n-rook:
I think this is a good idea. The underrated best part about using an automatic code formatter is that it actually makes formatting code less annoying; black just does things automatically and you don't get push failures due to bad formatting.

I think it's fine to format the migrations files but 32th is definitely right that black shouldn't run on the generated ks python files.

If we are going to use black we should probably
1. add a pre-commit git hook in the `git-hooks` directory
2. add a CI check for black

32th-System:
also, it seems like black and flake8 have conflicting opinions on code style. Which one do you want?

jgdhs27:
> also, it seems like black and flake8 have conflicting opinions on code style. Which one do you want?

I was surprised. The conflict is `comp_data[68 : 68 + comp_size]` where flake8 wants `comp_data[68:68 + comp_size]`.
As for which one I want, black.

jgdhs27:
> > also, it seems like black and flake8 have conflicting opinions on code style. Which one do you want?
> 
> I was surprised. The conflict is `comp_data[68 : 68 + comp_size]` where flake8 wants `comp_data[68:68 + comp_size]`. As for which one I want, black.

flake8 is a linter and black is an opinionated formatter, they are different kinds of tools. The proposal is to use a formatter instead of a linter.

n-rook:
I don't think we would want to get rid of flake8 just because we are using black. There are a few things that a formatter can't deal with but a linter can. Instead we would want to configure flake8 and turn off any linter check that disagrees with black (https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#flake8).

We'd also want a CI check to make sure that a PR ran black.

jgdhs27:
Excluded the folders that were suggested. Also, fixed compatibility with flake8, by ignoring E203 and setting max-line-length = 88 according to the [recommendations](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#flake8)

Idk how to configure git stuff. I tried adding a github action, but idk how to check if it works.

n-rook:
I'm ready to approve, but the one last thing I realized we need is a line in requirements-dev.txt. Otherwise different people might be using different versions of black and it would become a mess

jgdhs27:
I rebased + updated requirements-dev.txt. Looks like thre are not many PRs in progress right now. Could I please get another look?

PR #275, opened 2023-02-21
Bugfix: Fix SA Marisa A power display

jgdhs27:
Currently, SA Marisa A's power is displayed as power/20, when it should be power/12.

To fix this, I introduced the parameter `shot_name` to `GetFormatPower`. I get the shot name via a call to `replay_instance.shot.GetName()` - I have no idea if this is correct or how to test that.

### DIFF
--- a/project/thscoreboard/replays/game_fields.py
+++ b/project/thscoreboard/replays/game_fields.py
@@ -2,8 +2,9 @@
 
 import copy
 from immutabledict import immutabledict
-from typing import Optional
+from typing import Iterable, Optional
 from . import game_ids
+from replays import models
 
 #   These table fields configure the display of the stage split columns
 #   These may differ from whether the replay file actually has the field or not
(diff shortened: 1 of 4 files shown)


raviddog (approved):
Nice catch

jgdhs27:
I noticed that I have to use `shot_name == "Marisa A"` and not `shot_name == "MarisaA"`. Does that mean it's using the translated string and will it break on different languages?

raviddog:
> I noticed that I have to use `shot_name == "Marisa A"` and not `shot_name == "MarisaA"`. Does that mean it's using the translated string and will it break on different languages?

GetName() does appear to be calling GetShotName() which calls the pgettext functions, so that might be an issue. Looking at the model, it seems like the shot_id field is what you want

jgdhs27:
> > I noticed that I have to use `shot_name == "Marisa A"` and not `shot_name == "MarisaA"`. Does that mean it's using the translated string and will it break on different languages?
> 
> GetName() does appear to be calling GetShotName() which calls the pgettext functions, so that might be an issue. Looking at the model, it seems like the shot_id field is what you want

Thanks, fixed it.

PR #279, opened 2023-02-22
Download replays according to ingame user replay filename format

32th-System:


project/thscoreboard/replays/models.py
```diff
@@ -306,29 +306,28 @@ def IsVisible(self, viewer: auth.get_user_model()):
...
-                gamecode=gamecode,
-                id=self.id
-            )
-        else:
-            return '{gamecode}_{user}_{id}.rpy'.format(
-                gamecode=gamecode,
-                user=self.user.username,
-                id=self.id,
-            )
+        def numberToBase(n: int, b: str) -> str:
```
jgdhs27, line 316: I found it very difficult to understand what this function does and what `b` is. I would prefer if it was just `numberToBase36`. ~~Also where are the unit tests that show that this code works~~

jgdhs27 (changes requested):

project/thscoreboard/replays/test_game_ids.py
```diff
@@ -82,3 +82,12 @@ def testNoBugRouteNamesForGames(self):
...
                     game_ids.GetRouteName(game_id=game.game_id, route_id=route.route_id)
                 )
+
+
+class ReplayIdTestCase(test.SimpleTestCase):
+    
+    def testMakeReplayId(self):
+        self.assertEqual(game_ids.MakeReplayId(1679615), 'zzzz')
+        self.assertEqual(game_ids.MakeReplayId(1679616), '10000')
+        self.assertEqual(game_ids.MakeReplayId(96), '002o')
```
jgdhs27, line 92: I would really prefer `self.assertEqual(to_base_36(3 * 36 + 1), '0031')`. As it stands, this test doesn't explain itself and I have no idea why it should pass.

project/thscoreboard/replays/test_game_ids.py
```diff
@@ -89,5 +89,5 @@ class ReplayIdTestCase(test.SimpleTestCase):
     def testMakeReplayId(self):
         self.assertEqual(game_ids.MakeReplayId(1679615), 'zzzz')
         self.assertEqual(game_ids.MakeReplayId(1679616), '10000')
-        self.assertEqual(game_ids.MakeReplayId(96), '002o')
+        self.assertEqual(game_ids.MakeReplayId(3 * 36 + 1), '0031')
```
jgdhs27, line 92: I would strongly prefer if the function was called `to_base_36`. It explains what the function does, instead of just the context it's called in. When you call it with `rpy_id = to_base_36(self.id)` inside `GetNiceFilename` it's already clear what the goal was.
It also would let you write tests. Currently these tests just looks like "random number maps to random string". When I read these asserts I have no idea why they should be true and why you're testing for it.
(also, [snake_case for python functions](https://peps.python.org/pep-0008/#function-and-variable-names))

32th-System: where do you think I should put the function?

jgdhs27: would fit fine in models.py imo

32th-System: Or I could leave it where it is and name it `MakeBase36ReplayId`

jgdhs27: that works

jgdhs27 (approved):

PR #281, opened 2023-02-23
prevent signups with names/emails of invited users

32th-System:
This change is already on the production site. I manually edited the files and reloaded.


n-rook (approved):
I'm not against this change, but does it actually have any real effect? We're already in open beta, so I'm not sure how many more InvitedUser rows we're really going to see.

32th-System:
there is one InvitedUser row in the database right now

PR #282, opened 2023-02-24
Update translations

32th-System:


n-rook (approved):
I feel like there's something still off about our translation workflow, but I'm not sure what it is

PR #286, opened 2023-03-04
Improve reanalyze_replay so it can fix shot and route.

### LINKED ISSUE #246: reanalyze should really set shot and route if appropriate
Currently it doesn't, but this would be nice. We already have a replay that's bugged to have the wrong shot type (see #238) so even though the bug is fixed, we should be able to fix the old replay too.

n-rook:
Fixes #246

### DIFF
--- a/project/thscoreboard/replays/constant_helpers.py
+++ b/project/thscoreboard/replays/constant_helpers.py
@@ -13,6 +13,11 @@ class ReplayConstantModels:
     shot: models.Shot
     route: Optional[models.Route]
 
+    def SetOnReplay(self, r: models.Replay):
+        """Set the shot and route on a Replay instance."""
+        r.shot = self.shot
+        r.route = self.route
+
 
 def GetModelInstancesForReplay(replay_info: replay_parsing.ReplayInfo) -> ReplayConstantModels:
     """Get the constant model instances related to this replay."""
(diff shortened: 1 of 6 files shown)


project/thscoreboard/replays/constant_helpers.py
```diff
@@ -13,6 +13,11 @@ class ReplayConstantModels:
     shot: models.Shot
     route: Optional[models.Route]
 
+    def SetOnReplay(self, r: models.Replay):
```
jgdhs27, line 16: Please change this to a function `SetFromReplayConstants` inside the `models.Replay` class. It would be consistent with `SetFromReplayInfo`, and it makes more sense to mutate a `models.Replay` within its own class than within a dataclass.
Also, I think you should set `r.game` here too. If there's a reason why that shouldn't be done, then make the behaviour explicit from the function name.

jgdhs27 (changes requested):

project/thscoreboard/replays/constant_helpers.py
n-rook: There is no `r.game`! Since each shot corresponds to a specific game, we don't store the game for a replay separately; we just use the shot reference to determine which game it corresponds to.

For the method, I had wanted to do that but had been worried about cyclic imports. But you're right that it makes more sense that way, so I've done it now.

jgdhs27 (approved):
damn cyclic imports ;_;

PR #287, opened 2023-03-09
prevent duplicate replays from being submitted

### LINKED ISSUE #77: Don't let the same person upload the same replay multiple times
If someone tries to upload the same replay twice, the most likely explanation is, "They forgot they already uploaded it." So we shouldn't let people do that.

If two different people upload the same replay, it of course, it doesn't belong to both of them, but that case is harder to handle (maybe the same person forgot their account and created a new one, etc.) so that's a different problem altogether.

raviddog:
fixes #77 

scenarios:
1. attempt to upload a duplicate replay > validation error on replays/upload
2. proceed to publish page on the same replay twice > submit one > attempt to submit second > i dont know how to add an error message to the form yet


raviddog:
wait i don't actually need the code for scenario 2 because the publish url wont exist anymore once the temp replay is gone

project/thscoreboard/replays/views/create_replay.py
```diff
@@ -32,6 +34,13 @@ def _HandleReplay(request, replay_bytes):
 
     May raise ValidationError.
     """
+    
+    # test if replay already exists, and return an error if so
+    hash = hashlib.sha256(replay_bytes).digest()
```
n-rook, line 39: We should only have the code `hashlib.sha256(replay_bytes).digest()` in one place in the database. (Well, two places, it's fine to repeat it in the migration file.) That way there's no chance of a bug later if someone forgets what the right hash function is.

I would probably put the function in models.py... but I'm not really sure that's the right place. It would be fine to put it in a new module instead. (I don't think any of the existing modules are really the right place for it.)

project/thscoreboard/replays/views/create_replay.py
```diff
@@ -99,6 +108,14 @@ def publish_replay(request, temp_replay_id):
 
     if request.method == 'POST':
         form = forms.PublishReplayForm(replay_info.game, request.POST)
+
+        # test if replay already exists, and return an error if so
+        hash = hashlib.sha256(temp_replay.replay).digest()
+        duplicate = models.ReplayFile.objects.filter(replay_hash=hash)
+        if duplicate.exists():
+            # I don't know a better way of doing this right now
```
n-rook, line 116: I think I actually have a good idea here. Take a look at the django docs:

https://docs.djangoproject.com/en/4.1/ref/forms/api/#django.forms.Form.non_field_errors

Apparently if you call [Form.add_error(None, "...")](https://docs.djangoproject.com/en/4.1/ref/forms/api/#django.forms.Form.add_error) then the error will be recorded on the form as a "non-field error". Then I think it'll show up as an error on the form; you can see `{{ form.non_field_errors }}` in publish.html.

project/thscoreboard/replays/views/create_replay.py
```diff
@@ -32,6 +34,13 @@ def _HandleReplay(request, replay_bytes):
 
     May raise ValidationError.
     """
+    
+    # test if replay already exists, and return an error if so
+    hash = hashlib.sha256(replay_bytes).digest()
+    duplicate = models.ReplayFile.objects.filter(replay_hash=hash)
```
n-rook, line 40: If you felt clever you could add this as a method on the ReplayFile manager, like the visible_to function that exists now. But you don't have to.

n-rook (changes requested):
Nice, this'll be really good to have

project/thscoreboard/replays/views/create_replay.py
raviddog: I'm thinking some sort of redirect actually is the corrent play here, because the pending replay no longer exists so the url the user is on is actually invalid. I'm just concerned the standard 404 doesn't explain what happened well enough.

project/thscoreboard/replays/views/create_replay.py
raviddog: I could, but every time a duplicate needs to be checked, we haven't actually created the ReplayFile object yet, so it wouldn't do anything currently.

project/thscoreboard/replays/constant_helpers.py
```diff
@@ -27,3 +28,13 @@ def GetModelInstancesForReplay(replay_info: replay_parsing.ReplayInfo) -> Replay
         shot=shot,
         route=route
     )
+
+
+def CheckReplayFileDuplicate(file):
+    hash = hashlib.sha256(file).digest()
```
n-rook, line 34: This is duplicated code; just call CalculateReplayFileHash() below

project/thscoreboard/replays/views/create_replay.py
n-rook: Oh, I see what you mean. In that case, I think what you really want is a totally different-looking page that's an error page, right? For that, you can create a totally new template to render (probably inheriting from the simple centered template base, since it doesn't have to be complicated), and just say exactly what you want there. I think it's better to not do a real URL redirect, since the appropriate URL for the page really is the "finish publishing" URL, we just want to render a different page there that says "this replay was already uploaded". This also makes it easy for you to put a link on that page to the actual replay.

n-rook (changes requested):

project/thscoreboard/replays/views/create_replay.py
raviddog: I think the 404 will have to do because the user can't even get to the code again coz the URL doesn't exist and I'm not really fond of the idea of leaving the row in the db even though the replay is already published just to forward requests.

n-rook (approved):

Issue #259, opened 2023-02-17
Better search/filtering


32th-System:
Right now searching on the site follows a hierarchy where you can filter per game per difficulty but you can't filter per game per shot. You can only make queries that are per game per difficulty into queries that are per game per difficulty per shot. I think that any filtering at least beyond per game should be done with a query string. As for the proper search UI, I think that should be part of a more general site redesign.

PR #288, opened 2023-03-16
Use checkboxes for navigation

jgdhs27:
Rewrites the replay_table code to allow a user to filter for shot and difficulty more conveniently.

image

This code is implemented in JS. Since there is now some complex JS logic, I've also added some Jest tests.

The following is not done yet but should probably be done soon:
* Add filters for HSiFS Seasons
* Possibly better filters for IN
* Reintroduce the replay date column, maybe more columns
* Styling

Related Issue: #259


raviddog:
You've lost the link to show replay details.
Also not as important, but the score isn't comma delimited anymore (we should consider using localisation for this somehow, I don't really know how that works)

project/thscoreboard/replays/views/index.py
```diff
@@ -17,10 +18,20 @@ def index(request):
...
         [:10]
     )
+    
+    recent_replays = (
+        models.Replay.objects
+        .visible_to(request.user)
+        .filter(category__in=[models.Category.REGULAR, models.Category.TAS])
+        .order_by('-created')
+        [:10]
+    )
```
n-rook, line 28: Isn't this identical to the `recent_uploads` query above?

project/thscoreboard/shared_content/static/js/replays/replayTable.js
```diff
@@ -0,0 +1,93 @@
...
+var activeFilters = {};
+
+// initialize table content
+populateTable(allReplays);
+
+
+function onClick(elm) {
+  const filterType = elm.getAttribute('filterType');
+  const value = elm.getAttribute('value');
+  console.log(value);
```
n-rook, line 17: Do we want this?

project/thscoreboard/shared_content/static/js/replays/replayTable.js
```diff
@@ -0,0 +1,93 @@
...
+  const filteredReplays = filterReplays(activeFilters, allReplays);
+  populateTable(filteredReplays);
+}
+
+function updateFilters(filters, filterType, value) {
+  if (filters[filterType] === undefined) {
+    filters[filterType] = [];
+  }
+  const indexOfValue = filters[filterType].indexOf(value);
+  if (indexOfValue == -1) {
```
n-rook, line 29: I think we should avoid `==`, since it can have surprising results

project/thscoreboard/shared_content/static/js/replays/replayTable.js
```diff
@@ -0,0 +1,93 @@
...
+}
+
+function populateTable(replays) {
+  const replayTableHtml = document.getElementById('replay-table');
+  const tbody = replayTableHtml.getElementsByTagName('tbody')[0];
+  
+  // Clear the existing rows from the table body
+  tbody.innerHTML = '';
+
+  for (let i = 0; i < replays.length; i++) {
```
n-rook, line 58: Can't this just be `for (const replay of replays)`? That seems simpler. (The same is true of the other for loop below.)

project/thscoreboard/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "devDependencies": {
+    "@types/jsdom": "^21.1.0",
+    "jest": "^29.5.0",
+    "jsdom": "^21.1.1"
+  }
+}
```
n-rook, line 7: Add instructions on how to run the JS tests in the base readme.

project/thscoreboard/replays/views/index.py
```diff
@@ -4,6 +4,7 @@
 from django.views.decorators import http as http_decorators
 
 from replays import models
+from replays.views.replay_list import convert_replays_to_json_string
```
n-rook, line 7: Importing functions from one view to another feels a bit unexpected to me. Can we put it in a shared library elsewhere? That way, we could write a test against it, too.

project/thscoreboard/replays/templates/replays/game_scoreboard.html
```diff
@@ -2,8 +2,30 @@
...
-    {% elif difficulty %}<h3>{{difficulty_name}}</h3>
-    {% endif %}
+
+    <div class="checkbox-list">
+        {% for difficulty_name in difficulties %}
+            <label><input
+                type="checkbox"
+                filterType="Difficulty"
+                value="{{difficulty_name}}"
+                onclick="onClick(this)"
```
n-rook, line 12: I think it's too confusing to call functions defined in replayTable.js from here, when replayTable.js is imported in replay_table.html. I'm not sure what the best alternative is, though. Do you have any ideas?

project/thscoreboard/replays/templates/replays/replay_table.html
n-rook, line 1: Do other pages which include `replay_table` still work properly?

project/thscoreboard/replays/views/replay_list.py
n-rook, line 1: This eliminates the previous shot- and route-specific pages. That's fine, but double-check that they aren't still linked anywhere else on the site.

n-rook (changes requested):
Thanks for doing this, and sorry it took me a couple days to review it--- I haven't been feeling well. It largely looks great, but it's a big change so I have a bunch of comments.

jgdhs27:
> You've lost the link to show replay details. Also not as important, but the score isn't comma delimited anymore (we should consider using localisation for this somehow, I don't really know how that works)

Added back the link & made text comma delimited. It's not clear t me if people from other regions would prefer other formats, especially considering that the games and royalflare both used the 111,111,111 form.

project/thscoreboard/thscoreboard/templates/base.html
```diff
@@ -13,6 +13,19 @@
     <meta property="og:site_name" content="Silent Selene" />
     <meta property="og:description" content="{% block embed_description %}Upload and view Touhou replays at Silent Selene, the modern replay hub.{% endblock %}" />
     <meta property="og:image" content="{% block embed_favicon %}{% static 'favicon.png' %}{% endblock %}" />
+
+    <style>
```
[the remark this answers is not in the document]
jgdhs27: done

project/thscoreboard/replays/views/index.py
jgdhs27: oops, fixed

project/thscoreboard/shared_content/static/js/replays/replayTable.js
jgdhs27: removed

project/thscoreboard/shared_content/static/js/replays/replayTable.js
jgdhs27: good spot, fixed (here + 1 other place)

project/thscoreboard/shared_content/static/js/replays/replayTable.js
jgdhs27: yeah that simplifies the code a lot 👍

project/thscoreboard/package.json
jgdhs27: Added the following:
Install jest, for example with `npm install jest --global`, and run `jest` to run js tests
Hope its enough

project/thscoreboard/replays/templates/replays/game_scoreboard.html
jgdhs27: I can't think of any other way to do it. And I know this isnt a good reason, but that's how it was also done in `replay_details.js`

project/thscoreboard/replays/views/index.py
jgdhs27: makes sense. i extracted it to a new file `replays/replays_to_json` and wrote a test

project/thscoreboard/replays/templates/replays/replay_table.html
jgdhs27: Good point. I didn't realize that I need to update the use page too. Keeping multiple tables looks non-trivial so for now I decided to just return 1, in the future it would be possible to split by game.
image

project/thscoreboard/replays/templates/replays/game_scoreboard.html
n-rook: yeah... it's something we should figure out but it's out of the scope of this PR

n-rook (approved):

PR #290, opened 2023-03-18
Remove private replays

jgdhs27:
There was a month-old comment that read `Private is deprecated; it will be removed soon`. I guess this was forgotten about. Therefore I removed it.


raviddog:
Do you need to generate migrations again? Error seems to be complaining that the models dont match the migration database state

jgdhs27:
> Do you need to generate migrations again? Error seems to be complaining that the models dont match the migration database state

I guess so. How do I do that?

raviddog:
`py manage.py makemigrations`

jgdhs27:
Thank you!

raviddog (approved):
Nice

PR #291, opened 2023-03-18
Remove documentation from models.py

jgdhs27:
About 30% of the comments were lies and 60% were stating the obvious (`Replay.score """The score of the replay."""`). It seems that nobody was updating this documentation when things were changed/added. I see no reason to believe they will be maintained in the future. Therefore I removed them.

### DIFF
--- a/project/thscoreboard/replays/models.py
+++ b/project/thscoreboard/replays/models.py
@@ -1,6 +1,3 @@
-"""Contains all of the models for the replay site."""
-
-
 import dataclasses
 import datetime
 from typing import Optional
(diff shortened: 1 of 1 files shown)


raviddog:
I'm not comfortable with this. If 30% of the comments are wrong, then they should be fixed. I do believe a lot of it is overcommented but that isn't reason to remove every single comment entirely.

jgdhs27:
> I'm not comfortable with this. If 30% of the comments are wrong, then they should be fixed. I do believe a lot of it is overcommented but that isn't reason to remove every single comment entirely.

They'll be wrong again in a month

raviddog:
You're definitely on point that noone has touched this documentation for a while. I think a lot of it, especially the ReplayStage fields are overdocumented and those seem to be the ones that would require constant touchups. We do have a separate implementation for what fields are used per game so those can all be cleaned up.

I think its worth having comments that are stating the obvious though. A lot of the short one-liner docstrings still contain some context information that can prevent confusion on what a model field is for. Even though we've tried to keep the field names short and accurate, its possible there's room for confusion that we haven't considered.

Rather than deleting it, we should review the docstrings and clean up anything that would require ongoing maintenance or has duplicated information that can be found or is asserted elsewhere.

n-rook:
I'm with raviddog on this one. It is true that some of the docs are wrong but that just means we should do a better job of updating them. For what it's worth, I think most of these fields will not change in meaning anytime soon--- it is not like ZUN is putting out new Touhou games every month. So if we get the docs right they should be right going forward. Database fields are hard to change, so docs will go stale less often than they do for functions.

My experience is that it's a nightmare working on projects that don't document their database fields, because unlike functions it can be really hard to figure out what the fields are supposed to mean without any context. And if code is submitted that is wrong about what fields are supposed to mean, it can be miserable to correct once wronginformation winds up in the database. It is true that some of the docs are kinda trivial and dumb, but specifically for the database I think it's best to err on the side of more documentation.

My experience is that the best doc strategy for software projects is kinda of a hierarchy of docs, where the broader the scope of access for something, the more important it is to document it right. The hierarchy goes like:
* Code inside functions
* Private functions and other module-scoped entities
* Public functions and other module-scoped entities
* APIs which talk outside of the server

And talking to the database falls under the last category.

jgdhs27:
ok, I will close the PR

PR #293, opened 2023-03-19
update documentation

### LINKED ISSUE #292: Update models.py documentation
A lot of documentation in `models.py` is outdated, particularly in the `ReplayStage` stage. These should be updated.

raviddog:
fixes #292 

if the parser comments are too much i can cut back on them, but I really like the idea of having the information right next to the code, even if it is duplicated


project/thscoreboard/replays/models.py
jgdhs27, line 1: The docstring of `def CleanUp(cls, now: datetime.datetime) -> None:` says that an int is returned, this needs to be updated too

project/thscoreboard/replays/replay_parsing.py
```diff
@@ -530,6 +544,7 @@ def _Parse12(rep_raw):
         if next_stage is not None:
             s.score = next_stage.score * 10
             s.power = next_stage.power
+            # piv is stored with extra precision, we trunctate the value to what is shown ingame
             s.piv = (math.trunc(next_stage.piv / 1000)) * 10
```
jgdhs27, line 548: imo this should be extracted to a function `def convert_piv_stored_in_replay_to_displayed_piv(game_id, PIV: int) -> int`

project/thscoreboard/replays/replay_parsing.py
```diff
@@ -112,11 +113,14 @@ def _Parse06(rep_raw):
 
     rep_stages = []
 
+    # TH06 stores stage data values from the start of the stage but score from the end
```
jgdhs27, line 116: This comment is too far from the code it's describing

project/thscoreboard/replays/replay_parsing.py
raviddog: this is looking awfully similar to the game_fields format functions, should we be storing the PIV the same as in the file, and just converting it visually along with the rest of the data there?

project/thscoreboard/replays/replay_parsing.py
```diff
@@ -427,6 +436,7 @@ def _Parse10(rep_raw):
 
     rep_stages = []
     
+    # TH10 stores all stage data values from the start of the stage so we need to retrieve them from the next stage
```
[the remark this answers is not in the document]
raviddog: good idea

project/thscoreboard/replays/replay_parsing.py
```diff
@@ -901,13 +933,16 @@ def _Parse18(rep_raw) -> ReplayInfo:
 
     rep_stages = []
     
+    # TH18 contains both stage start and end data for each stage
+    # However the end-of-stage data is inaccurate as explained below, so the next stage data is still required
```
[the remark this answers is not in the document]
raviddog: My thought process was the code these comments are on are also linked anyway, but you're right this isn't needed

project/thscoreboard/replays/models.py
```diff
@@ -329,9 +329,8 @@ def SetForeignKeysFromConstantModels(self, c: ReplayConstantModels):
 class ReplayStage(models.Model):
     """ Represents the end-of-stage data for a stage split for a given replay
         The data may not directly correspond to how it is stored in-game, since some games store it differently
-        Most games store the values from the start of the stage
-        TH07 stores the values from the end of the stage
-        TH08 is weird in that the score is stored from end of stage, but everything else is from the start
+        All differences/quirks in how this data is stored in-game is handled in the relevant replay_parsing.Parse() functions
```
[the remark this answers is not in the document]
raviddog: This goes along with the PIV function comment, currently there's 2 places where data is getting transformed from the replay file, and its not consistent whether fields in the DB are pre- or post transform. Documenting them in the models would eventually be forgotten as you say, but we need some form of either consistency in the data, or somewhere to say what the ruling on how data gets entered into each field is. I'm not sure which path we should take.

raviddog:
I know we don't agree on a lot of this but I do think this is an improvement over what's currently in the codebase.  You can reopen the issue if you feel further changes are warranted but lets get this moving forward a bit.

jgdhs27 (approved):

PR #294, opened 2023-03-19
Add NB and NM annotations to replays.

n-rook:
#78

image

image

This PR adds two new pieces of user-selected information. There's a "bombs" checkbox, and an optional "miss count" text box.

I'm not super happy with the user experience, but the "publish replay" and "view replay" pages both need a holistic redesign anyway, and I figured it'd be easier to do that once we have this information.


project/thscoreboard/replays/game_ids.py
```diff
@@ -506,3 +508,42 @@ def MakeBase36ReplayId(id: int) -> str:
...
+        game_id: The game.
+        replay_type: The type of replay. If unset, this function returns True
+            if any replay type (for the given game) has bombs.
+    """
+    if game_id == GameIDs.TH09:
+        # PoFV does not have traditional bombs.
+        return False
+
+    if replay_type == ReplayTypes.SPELL_PRACTICE:
+        # You can't bomb in spell practice.
```
32th-System, line 526: Are you sure about that?

project/thscoreboard/replays/game_ids.py
```diff
@@ -506,3 +508,42 @@ def MakeBase36ReplayId(id: int) -> str:
...
+        replay_type: The type of replay. If unset, this function returns True
+            if any replay type (for the given game) has bombs.
+    """
+
+    # All currently supported games have lives, but scene games don't, so in
+    # the future we will return False sometimes here.
+    del game_id  # Stop flake8 from complaining that it is unused.
+
+    if replay_type == ReplayTypes.SPELL_PRACTICE:
+        # You can't die in spell practice.
```
32th-System, line 546: Are you sure about that?

project/thscoreboard/replays/replay_parsing.py
```diff
@@ -50,10 +50,10 @@ class ReplayStage:
...
     point_items: int = None
     power: int = None
-    lives: int = None
-    life_pieces: int = None
-    bombs: int = None
-    bomb_pieces: int = None
+    lives: Optional[int] = None
+    life_pieces: Optional[int] = None
+    bombs: Optional[int] = None
+    bomb_pieces: Optional[int] = None
```
jgdhs27, line 56: now that im looking at it, shouldnt all the type hints here be optional?

jgdhs27 (approved):
As a user I have a strong opinion that user-specified fields should be optional. I assume that would be a part of the "holistic redesign"

project/thscoreboard/replays/replay_parsing.py
32th-System: I tried doing some typechecking on these and yes, all of these fields except stage number should be optional

n-rook:
In theory I agree that fields like this should be optional, but I couldn't think of any reasonable UI that would let "no bomb" be optional without adding complexity to the input. Something like a drop-down with "yes, no, unknown" or a ternary checkbox is way harder to understand than a regular checkbox. And it isn't a huge deal anyway, because in general, if the player doesn't care about NB, then they used bombs.

project/thscoreboard/replays/replay_parsing.py
jgdhs27: even stage number, the line `stage: int = None` is incorrectly typed

32th-System: in my typechecking experiments I just didn't initialize the stage number in the struct and would instead do it when making an instance of the class

project/thscoreboard/replays/game_ids.py
n-rook: I think it's true, but even if not, I don't think it would make sense to put "nb" or "miss count" inputs on spell practice replays

32th-System: On some spells in th16 you can get enough points to get the first score extend in spell practice, allowing you to die and respawn with 3 bombs.

n-rook: How bizarre. I fixed the comments to be more accurate, since given this they were wrong. I think it's still good to hide these UI elements for spell practice, since this is a rare case.

project/thscoreboard/replays/game_ids.py
n-rook: See above

project/thscoreboard/replays/replay_parsing.py
n-rook: I think you're right. These wound up sneaking into this PR by accident; I had a follow-up where I used them (and was confused by them being optional), but it isn't relevant to this PR. But I suppose I might as well commit this part anyway, since it _is_ correct.

PR #304, opened 2023-03-29
Import replays from royalflare

jgdhs27:
Creates a new script called import_royalflare, which imports replays from royalflare. For instructions for how to use it, run `manage.py import_royalflare --help`.

To accommodate replays without a user, I made the "replay.user" field optional, and added a new field "replay.imported_username".


project/thscoreboard/replays/management/commands/import_royalflare.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def handle(self, *args, **options):
+        print(ROYALFLARE_JSON_DIRECTORY_PATH)
+        main(options["replay_dir"])
+
+
+def main(replay_dir: str) -> None:
+    json_files = ROYALFLARE_JSON_DIRECTORY_PATH.glob("**/*.json")
+    for json_file in json_files:
+        with open(json_file, encoding="utf-8") as f:
```
32th-System, line 42: I think it would be better to specify files on the command line, instead of telling it to look at all files in an entire folder.

32th-System (changes requested):
I would recommend a constraint on the Replay model where the condition is basically ` imported_username__isnull=True ^ user__isnull=True` (^ is xor).

32th-System:
Update: here's what the constraint looks like in Django code:
```python
            models.CheckConstraint(
                name="user_imported_username_xor_isnull",
                check=(
                    models.Q(user__isnull=True, imported_username__isnull=False)
                    | models.Q(user__isnull=False, imported_username__isnull=True)
                )
            ),
```

project/thscoreboard/replays/management/commands/import_royalflare.py
jgdhs27: added it as an optional argument

jgdhs27:
> I would recommend a constraint on the Replay model where the condition is basically ` imported_username__isnull=True ^ user__isnull=True` (^ is xor).

i agree with a need for a constraint, but idk if we should disallow both fields 

> Collaborator

added

32th-System:
constraints are things enforced by the database itself and require a `makemigrations` and a `migrate`

jgdhs27:
> constraints are things enforced by the database itself and require a `makemigrations` and a `migrate`

fixed 😅

project/thscoreboard/replays/create_replay.py
```diff
@@ -27,6 +28,8 @@ def PublishNewReplay(
     miss_count: Optional[int],
     temp_replay_instance: models.TemporaryReplayFile,
     replay_info: replay_parsing.ReplayInfo,
+    created_timestamp: Optional[datetime.datetime] = None,
+    imported_username: Optional[str] = None,
```
n-rook, line 32: Please update the docstring as well

project/thscoreboard/replays/replays_to_json.py
```diff
@@ -11,7 +11,9 @@ def convert_replays_to_serializable_list(
             "User": {
                 "text": f"{replay.user.username}",
                 "url": f"/replays/user/{replay.user.username}",
-            },
+            }
+            if replay.user
+            else replay.imported_username or replay.name,
```
n-rook, line 16: I think it's too much to expect from the frontend for it to handle a "User" json field that sometimes is an object and sometimes is a string. I'd rather have return "text" as the username in question and "url" be empty, and handle it appropriately from there.

n-rook (commented):
This largely looks good. Submitting my review as a comment so you don't have to wait for my approval too.

We should think about how we want to display RF replays on the site, and how we should handle the possibility that a user will want to delete their old replays.

project/thscoreboard/replays/create_replay.py
jgdhs27: done

project/thscoreboard/replays/replays_to_json.py
jgdhs27: I don't necessarily agree. for all fields, the frontend expects to get either a `string`, or a map `{'text' => string, 'url' => string}`. handling this doesn't add any logic to the frontend.

32th-System (changes requested):
Everything is fine, but I realized that you should also prettify the JSON files (or not have them at all)

jgdhs27:
> Everything is fine, but I realized that you should also prettify the JSON files (or not have them at all)

done

32th-System (approved):

PR #306, opened 2023-03-31
Redesign the site layout, front page, game pages

32th-System:
Design made by Nylilsa. This is just stage 1 of the redesign. Pretty much all pages will need adapting. In it's current state the site is functional, but only the front page is pretty.


32th-System:
Current state:
https://www.youtube.com/watch?v=iTeCeMh4Tg0

project/thscoreboard/thscoreboard/templates/base.html
```diff
@@ -1,63 +1,144 @@
...
 <html lang="en">
 <head>
     <title>{% block title %}Silent Selene{% endblock %}</title>
-    <link rel="stylesheet" href="{% sass_src 'style.sass' %}">
     <link rel="stylesheet" href="{% static 'fonts/Lato.css' %}">
     <link rel="stylesheet" href="{% static 'fonts/EBGaramond.css' %}">
+    <link rel="stylesheet" href="{% static 'style.css' %}">
+    <script src="/static/js/main.js" defer></script>
     {% block favicon %}<link rel="icon" type="image/png" href="{% static 'favicon.png' %}" sizes="48x48">{% endblock %}
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
```
n-rook, line 12: afaik this isn't relevant, see https://stackoverflow.com/questions/26346917/why-use-x-ua-compatible-ie-edge-anymore

project/thscoreboard/shared_content/static/style.css
```diff
@@ -0,0 +1,287 @@
+*,
```
n-rook, line 1: I've mentioned this in the discord before, but I'm against deleting the existing style sheets. I think we will lose a lot of formatting by accident and it makes it impossible to review this and tell what's actually changed. If you want, I can take a look tonight and tomorrow at changing this PR to do that.

n-rook (changes requested):

project/thscoreboard/replays/templates/replays/index.html
```diff
@@ -2,20 +2,10 @@
...
-        </li>
-    </ul>
-    {% endfor %}
-</p>
-
-<p>Recent uploads:
-    {% include "replays/replay_table.html" with show_game=True %}
-</p>
+<p class="highlight">Welcome to Silent Selene !</p>
+<p>Lorem ipsum dolor sit amet consectetur adipisicing elit. Maxime tenetur cum voluptatum tempore atque optio consequuntur quos, quas, sint, recusandae eligendi odio. Animi consequatur placeat nemo doloremque quas illum id enim aspernatur aliquam soluta, maiores reiciendis! Sed excepturi odio consectetur?</p>
```
jgdhs27, line 6: we still need to figure out what to put here

project/thscoreboard/replays/test_replays_to_json.py
```diff
@@ -60,7 +61,8 @@ def testConvertReplaysToJsonString(self):
         replays = [replay_1, replay_2]
 
         converter = ReplayToJsonConverter()
-        json_data = converter.convert_replays_to_serializable_list(replays)
+        with override_settings(LANGUAGE_CODE="pt-br"):
```
jgdhs27, line 64: what?

32th-System: The function, because it uses gettext, doesn't always have to output the same thing. The test expects "th06", but if `compilemessages` was run, the function returns "EoSD" instead. To make sure that gettext doesn't affect anything, the function is run with a language we don't support. In another test that encounters issues with gettext, we use Brazilian Portugese, so I also use Brazilian Portugese here.

jgdhs27: i see. in that case, i think it makes more sense to call `json_data = converter.convert_replays_to_serializable_list(replays)` in a real locale (eg. English) and check that it's converted correctly. After all, this is sent and to frontend and displayed, we want to check its display value is good.

jgdhs27: i just saw the comment on the other usage and maybe that doesnt work in github CI? if thats the case, then please at least add a comment explaining that

Issue #309, opened 2023-03-31
Bad performance of replay tables


jgdhs27:
With 1000s of replays imported you big to notice some performance issues. I have 4656 entries for Imperishable Night, and it takes 2.8 seconds for the table to load.

Applying filters can also have bad performance. The performance depends on the number of rows in the resulting table.

Some quick profiling:

For the first load:
- in total, the page takes 2.79s to load
- 1.85s on the network request for replays/th08/json
- 87ms on xhr.onload
  - 50ms is spent in populateTable()
- 663ms rendering the table
  - 132ms on recalculating style
  - 420ms on page layout
  - 69ms on pre-paint

Toggling off a filter so that all rows are show again:
- 122ms on onClick()
  - 122ms spent in populateTable()
- 706ms rendering the table
  - 153ms on recalculating style
  - 391ms on page layout
  - 73ms on pre-paint

jgdhs27:
My proposal would be to split both the API call and the table rendering into two stages. The first stage would be limited to 20 replays, then the second stage would process all replays. Also, all rendering should be moved to a new thread to not block the UI.

jgdhs27:
Doesn't look like its possible to render on a separate thread

jgdhs27:
I found a solution to the table rendering bottleneck. Create several tables, one under another, separated by divs. Only the first one gets a header. Then load them in series. This makes the rendering feel fast (because the top rows that you see belong to the first table) and only blocks the main thread for as long as it takes to render 1 smaller table.

I'm still wondering if there's a smarter way to do things.

n-rook:
This was my main concern with client-side rendering and filtering of tables.

Loading them in series makes sense to me.

PR #312, opened 2023-04-03
Split up replay table

jgdhs27:
Currently, the performance of rendering a large table is bad. For Th08, where I have 4656 replays, it takes quite long for a table to render. During this time, the page does not update and the UI is unresponsive.

To mitigate this, a large replay table is split up into many smaller replay tables that are loaded in series. For every 50 rows, a new table is created. This reduces the time for the page to update, and freezes the UI for less long.

| |Before | After | 
|---|---|---|
| Time taken for page to update (pointerup -> first table rendered) | 0.75s | 0.02s |
| Time taken for page to fully render (pointerup -> last table rendered) | 0.75s  | 0.73s |
| Total blocking time | 0.65s  | 0.33s |

Related issue: #309

### DIFF
--- a/project/thscoreboard/replays/templates/replays/replay_table.html
+++ b/project/thscoreboard/replays/templates/replays/replay_table.html
@@ -1,5 +1,5 @@
 {% load humanize %}
-<table id="replay-table" class="replay-table">
+<table class="replay-table">
     <thead>
         <tr>
             <th>
@@ -27,9 +27,10 @@
             </th>
         </tr>
     </thead>
-    <tbody>
+    <tbody id="replay-table-body">
     </tbody>
 </table>
+
 
 <!-- This script is a hack to get variables from python into a js file  -->
 <script>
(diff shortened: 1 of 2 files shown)


jgdhs27:
one thing i forgot to mention is that the table column widths are broken in this pr. it is required that table columns have fixed widths. however, any changes i make to styling will just clash with #306, so I am putting that off for now.

32th-System:
Actually, changing layout things is fine because everything related to layout would get completely overwritten by #306

raviddog:
Can we use pagination instead of having an everscroll table or has this already been discussed?

32th-System:
a really big table might be an issue on phones but on desktop it isn't too bad and you can Ctrl + F more easily

raviddog:
> a really big table might be an issue on phones but on desktop it isn't too bad and you can Ctrl + F more easily

It's more just I hate when you navigate away from the page and then come back and you have to scroll down 50 times to get back to where you were.

n-rook (approved):
Approved because this is clearly an improvement over status quo

PR #314, opened 2023-04-04
Cache expensive operations in replay to json conversion

jgdhs27:
In #309 the network request for the json API was reported to be extremely slow. I looked into it, and the reason was that the funciton `convert_replays_to_serializable_list` was slow. This function makes thousands of calls like `game.GetShortName()` which appear to be expensive (presumably, because they need to fetch a user's locale and create a translated string?).

Since many of these calls are the same (for instance, `game.GetShortName()` will return the same value every time for a game page), I cached all expensive calls. I implemented this using `functools.lru_cache`. This brought the time taken to execute the function for 4500 replays down from 1.7 seconds to 0.2 seconds.

| | Before | After |
| --- | --- | ----------- |
| convert_replays_to_serializable_list | 1.7s | 0.2s |
| th08/json network request| 1.8s | 219ms |
| time for table to display after page load (with #312 checked out) | 1.99s | 589ms |

~~Even after these changes, the function `convert_replays_to_serializable_list` remains one of the most expensive parts of loading the page.~~ Edit: Actually, after the changes, `convert_replays_to_serializable_list` is not really a bottleneck. Even if it was instant, the th08/json network request still takes about 200ms. Unsure what the new bottleneck is.

### DIFF
--- a/project/thscoreboard/replays/replays_to_json.py
+++ b/project/thscoreboard/replays/replays_to_json.py
@@ -9,14 +9,6 @@ class ReplayToJsonConverter:
         return shot.game
 
     @lru_cache(maxsize=None)
-    def _get_game_short_name(self, game: models.Game) -> str:
-        return game.GetShortName()
-
-    @lru_cache(maxsize=None)
-    def _get_difficulty_name(self, game: models.Game, replay: models.Replay) -> str:
-        return game_ids.GetDifficultyName(game.game_id, replay.difficulty)
-
-    @lru_cache(maxsize=None)
     def _get_shot_name(self, shot: models.Shot) -> str:
         return shot.GetName()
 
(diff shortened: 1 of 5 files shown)


project/thscoreboard/replays/replays_to_json.py
```diff
@@ -1,32 +1,45 @@
...
-        for replay in replays
-    ]
-    return replay_dicts
+
+    def convert_replays_to_serializable_list(
+        self,
+        replays: Iterable[models.Replay],
+    ) -> list[dict[str, any]]:
+        a = [self._convert_replay_to_dict(replay) for replay in replays]
+        return a
```
32th-System, line 45: Does `return [self._convert_replay_to_dict(replay) for replay in replays]` not work? Does it make the function slow?

jgdhs27: oops! fixed

32th-System (approved):

PR #318, opened 2023-04-06
Add comments to replay tables

jgdhs27:
Adds comments to replay tables. If the comment contains more than 250 characters, it is cut off and only the first 250 characters are shown. Also adds some small formatting changes to the table, to prevent certain entries (shot name, date) from wrapping and taking up muiltiple lines.

image

### DIFF
--- a/project/thscoreboard/replays/models.py
+++ b/project/thscoreboard/replays/models.py
@@ -360,6 +360,13 @@ class Replay(models.Model):
         self.shot = c.shot
         self.route = c.route
 
+    def GetShortenedComment(self) -> str:
+        """Gets a short version of the replay comment, suitable to be displayed
+        in a table row."""
+        if len(self.comment) <= limits.MAX_SHORTENED_COMMENT_LENGTH:
+            return self.comment
+        return self.comment[:limits.MAX_SHORTENED_COMMENT_LENGTH] + "..."
+
 
 class ReplayStage(models.Model):
     """Represents the end-of-stage data for a stage split for a given replay
--- a/project/thscoreboard/shared_content/static/js/replays/replayTable.js
+++ b/project/thscoreboard/shared_content/static/js/replays/replayTable.js
@@ -102,6 +102,9 @@ function createTableCell(columnName, value) {
     const text = document.createTextNode(value);
     cell.appendChild(text);
   }
+  if (columnName === "Shot") {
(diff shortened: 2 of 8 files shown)


project/thscoreboard/shared_content/static/js/replays/replayTable.js
```diff
@@ -102,6 +102,9 @@ function createTableCell(columnName, value) {
     const text = document.createTextNode(value);
     cell.appendChild(text);
   }
+  if (columnName === "Shot") {
+    cell.style.whiteSpace = "nowrap";
```
n-rook, line 106: It's easier to understand HTML and make extra changes if we keep as much as possible to CSS. As such, we should put the "white-space: nowrap" in CSS, and here we should add a class to activate that CSS.

project/thscoreboard/replays/models.py
```diff
@@ -360,6 +360,13 @@ def SetForeignKeysFromConstantModels(self, c: ReplayConstantModels):
         self.shot = c.shot
         self.route = c.route
 
+    def GetShortenedComment(self) -> str:
+        """Gets a short version of the replay comment, suitable to be displayed
+        in a table row."""
+        if len(self.comment) <= limits.MAX_SHORTENED_COMMENT_LENGTH:
+            return self.comment
+        return self.comment[:limits.MAX_SHORTENED_COMMENT_LENGTH] + "..."
```
n-rook, line 368: We need something like this, but in addition I think we should also still use "text-overflow: ellipsis" on the frontend. That'll be good for places like mobile where users likely have small screens.

n-rook (changes requested):

project/thscoreboard/shared_content/static/js/replays/replayTable.js
jgdhs27: Added a new style  `.nowrap white-space: nowrap;` to `style.sass`

project/thscoreboard/replays/models.py
jgdhs27: I actually want the text to be able to take up a couple of lines if needed

n-rook: Fair enough! I'm not sure about this, but I don't feel strongly enough to gainsay you about it.

n-rook (approved):

PR #320, opened 2023-04-06
Derive replay type from game

### LINKED ISSUE #316: Other PC-98 games incorrectly displaying "PVP" as a replay type option
image
image
LLS and MS seem to have this too. However, only PoDD has a PvP mode.

jgdhs27:
Derive the replay type choices in the "Publish Replay Without File Form" from the game.

Currently, replays uploaded without file will always show all replay types, including PVP. This is a bug, as TH01, TH02, TH04 and TH05 do not support PVP.

Fixes #316

### DIFF
--- a/project/thscoreboard/replays/forms.py
+++ b/project/thscoreboard/replays/forms.py
@@ -61,6 +61,17 @@ def _create_comment_field():
     )
 
 
+def _get_replay_type_choices(game: models.Game) -> list[Tuple[str, str]]:
+    replay_types = [
+        ("1", _("Regular")),
+        ("2", _("Stage Practice")),
+    ]
+    games_with_pvp = [game_ids.GameIDs.TH03, game_ids.GameIDs.TH09]
+    if game.game_id in games_with_pvp:
+        replay_types.append(("4", _("PVP")))
+    return replay_types
+
+
 class ShotField(forms.ModelChoiceField):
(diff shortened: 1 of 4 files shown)


jgdhs27:
Not sure why the test is failing. it passes with `python3.10 manage.py test -p test_forms.py`, but fails with `python3.10 manage.py test`.

project/thscoreboard/replays/forms.py
```diff
@@ -55,6 +55,17 @@ def _GetDifficultyChoices(game: models.Game) -> Tuple[int, str]:
     ]
 
 
+def _get_replay_type_choices(game: models.Game) -> list[Tuple[str, str]]:
+    replay_types = [
+        ("1", _("Regular")),
+        ("2", _("Stage Practice")),
+    ]
+    games_with_pvp = [game_ids.GameIDs.TH03, game_ids.GameIDs.TH09]
```
raviddog, line 63: I would recommend moving this array to game_fields.py

raviddog (changes requested):

project/thscoreboard/replays/forms.py
jgdhs27: done. although the data still feels a tiny bit out of place there.

jgdhs27:
rebased onto main

n-rook (approved):

raviddog (approved):

PR #323, opened 2023-04-07
Fix replay username parsing

### LINKED ISSUE #322: Error thrown when parsing username from replay
In the relay parsers, the following error is thrown on certain replays:
`'ascii' codec can't decode byte 0xa5 in position 3: ordinal not in range(128)`

Guess: this happens when the username in the replay contains a space.

Example of a failing replay: http://replay.lunarcast.net/replay/th6_ud5071.rpy

jgdhs27:
Usernames are encoded using SJIS encoding. However, the parsers used ASCII encoding. This led to errors being thrown on replays that had these characters in their username.

Fixes #322


32th-System (approved):

raviddog:
I'll submit my PR for #324 once this gets merged

PR #325, opened 2023-04-08
Th modern fix royalflare corrupt userdata

### LINKED ISSUE #324: 20% of Royalflare imports failing due to corrupted replay headers
Many royalflare replays cannot be imported due to their headers being corrupted. For example, http://replay.lunarcast.net/replay/th11_ud5072.rpy About 20% of royalflare replays from th10 onwards are affected. Because the amount of affected replays is so high, we should figure out how to import them (and possibly even fix them). The replays work fine in-game.

raviddog:
fixes #324 

This change treats the user description section of the userdata as just raw bytes instead of any string encoding.
I'm worried about the byte signifying the difference between TH13/TH14 changing more than I've discovered with the replays I was provided, but I can't do anything outside of wait for those replays to show up and then alter this code further to fix it.


project/thscoreboard/replays/test_replay_parsing.py
```diff
@@ -266,6 +266,9 @@ def testSubshot(self):
     def testSmallFile(self):
         ParseTestReplay("th11_small_file")
 
+    def testRoyalflare(self):
+        ParseTestReplay("th11_royalflare")
```
jgdhs27, line 270: please rename the test to `testRoyalflareCorrupted`, and similar for the test files. otherwise i have no idea what this is

jgdhs27 (changes requested):
Thank you. This fixes almost all issues, but there are still ten th14 replays that are affected. Example: http://replay.lunarcast.net/replay/th13_ud5075.rpy . It would be nice if you could look into this one too. (and yes thats a th14 replay, lunarcast misidentified the game)

jgdhs27 (approved):
it works :D

PR #328, opened 2023-04-12
Support th16 season as a replay table filter

### LINKED ISSUE #295: Improve HSiFS filters
Currently, we only support filtering by difficulty and shot. This creates a mess in HSiFS, because there are 20 possible shots.

Suggestion: add a filter by season for HSiFS only.

jgdhs27:
Allow the replay table to filter th16 replays by character and by season independently.

image

After this is approved we can try doing something similar for TH17 subshot, and maybe even TH08 route.

Fixes #295

### DIFF
--- a/project/thscoreboard/replays/models.py
+++ b/project/thscoreboard/replays/models.py
@@ -87,6 +87,14 @@ class Shot(models.Model):
         """Get a pretty name for this shot type. Note: Populates game."""
         return game_ids.GetShotName(self.game.game_id, self.shot_id)
 
+    def GetCharacterame(self) -> str:
+        """Get a pretty name for this Character. Note: Populates game."""
+        return game_ids.GetCharacterName(self.game.game_id, self.shot_id)
+
+    def GetSubshotName(self) -> str:
+        """Get a pretty name for this subshot. Note: Populates game."""
+        return game_ids.GetSubshotName(self.game.game_id, self.shot_id)
+
 
 class Category(models.IntegerChoices):
     """The category under which a replay is uploaded."""
@@ -365,13 +373,13 @@ class Replay(models.Model):
         in a table row."""
         if len(self.comment) <= limits.MAX_SHORTENED_COMMENT_LENGTH:
             return self.comment
-        return self.comment[:limits.MAX_SHORTENED_COMMENT_LENGTH] + "..."
+        return self.comment[: limits.MAX_SHORTENED_COMMENT_LENGTH] + "..."
 
 
 class ReplayStage(models.Model):
-    """ Represents the end-of-stage data for a stage split for a given replay
-        The data may not directly correspond to how it is stored in-game, since some games store it differently
-        Many games only store the data from the start of a replay, so many of the fields for the final stage will be null
+    """Represents the end-of-stage data for a stage split for a given replay
+    The data may not directly correspond to how it is stored in-game, since some games store it differently
+    Many games only store the data from the start of a replay, so many of the fields for the final stage will be null
     """
 
     class Meta:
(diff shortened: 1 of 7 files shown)


project/thscoreboard/replays/models.py
```diff
@@ -69,7 +69,9 @@ class Shot(models.Model):
 
     class Meta:
         constraints = [
-            models.UniqueConstraint("shot_id", "game", name="unique_shot_per_game")
+            models.UniqueConstraint(
+                "shot_id", "game", "th16_season_id", name="unique_shot_per_game"
+            )
         ]
 
     shot_id = models.TextField()
```
n-rook, line 77: Hmm... I don't really like the idea of changing the database schema so that shot_id is no longer unique across a game. It seems like a big change to make, and a nice thing about the previous way of doing things was that it gave a unique name that roughly corresponded to how people talked about the shots. I'd rather see us add a season field, but retain the unique shot IDs, even if it's a little redundant.

project/thscoreboard/replays/models.py
```diff
@@ -83,9 +85,26 @@ class Meta:
     game = models.ForeignKey("Game", on_delete=models.CASCADE)
     """The game in which this shot appears."""
 
+    th16_season_id = models.TextField(blank=True, null=True)
```
n-rook, line 88: This concept is actually generic, IMO: I think we should use the same field to deal with th16 seasons and th17 animals. As such, I think a more generic field name is appropriate here; you don't actually have to implement it for wbawc in this PR, but we should leave the possibility open without further schema changes.

project/thscoreboard/replays/management/commands/setup_constant_tables.py
```diff
@@ -86,30 +86,22 @@ def _Create16():
     th16.save()
 
     shots = [
```
n-rook, line 88: This will cause issues in production, because _Create16() won't run. Unfortunately I think we have to add an update function in setup_constant_tables to do this right.

project/thscoreboard/replays/test_replays_to_json.py
```diff
@@ -66,18 +68,6 @@ def testConvertReplaysToJsonString(self):
 
         for replay, json_replay_data in zip(replays, json_data):
             assert json_replay_data
-            assert set(json_replay_data.keys()) == {
```
n-rook, line 69: Why don't you think this assertion is necessary anymore?

n-rook (changes requested):

project/thscoreboard/replays/models.py
jgdhs27: Godd point. I was thinking about this and we actually need to get a season and character for each shot for the frontend. (More generic: character + subshot). And while that could be implemented as two new redundant fields, I'm starting to think it makes more sense to just calculate them. I will try this approach.

project/thscoreboard/replays/test_replays_to_json.py
jgdhs27: All keys are checked by value anyway

jgdhs27:
Rewrote a few things so that the character and subshot, which are used in the frontend only, are now computed from the shot. This makes the change quite a bit smaller, and easier to adapt to other games (eg. TH17 with its subshots)

project/thscoreboard/replays/management/commands/setup_constant_tables.py
jgdhs27: The newer implementation doesn't require any migrations

project/thscoreboard/replays/models.py
```diff
@@ -87,6 +87,14 @@ def GetName(self):
         """Get a pretty name for this shot type. Note: Populates game."""
         return game_ids.GetShotName(self.game.game_id, self.shot_id)
 
+    def GetCharacterame(self) -> str:
```
32th-System, line 90: It seems like you typoed this name but you then just kept using the typoed name everywhere?

jgdhs27: fixed

n-rook (approved):
Yeah, doing it before the database makes sense to me. At the very least, it has the positive effect of making it so that we don't have to worry about migrations.

PR #331, opened 2023-04-15
Display route & filter by route in TH01/TH08

jgdhs27:
image

### DIFF
--- a/project/thscoreboard/replays/replays_to_json.py
+++ b/project/thscoreboard/replays/replays_to_json.py
@@ -15,31 +15,36 @@ class ReplayToJsonConverter:
     def _convert_replay_to_dict(self, replay: models.Replay) -> dict:
         shot = replay.shot
         game = self._get_game(shot)
-        json_dict = {
-            "Id": replay.id,
-            "User": {
+        include_route = game.game_id == game_ids.GameIDs.TH08
+        if include_route:
+            route = replay.route
+
+        json_dict = {}
+        json_dict["Id"] = replay.id
+        if replay.user:
+            json_dict["User"] = {
                 "text": f"{replay.user.username}",
                 "url": f"/replays/user/{replay.user.username}",
             }
-            if replay.user
-            else replay.imported_username or replay.name,
-            "Game": game.GetShortName(),
-            "Difficulty": game_ids.GetDifficultyName(game.game_id, replay.difficulty),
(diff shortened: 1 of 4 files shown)


project/thscoreboard/replays/replays_to_json.py
```diff
@@ -15,27 +15,35 @@ def _get_shot_name(self, shot: models.Shot) -> str:
     def _convert_replay_to_dict(self, replay: models.Replay) -> dict:
         shot = replay.shot
         game = self._get_game(shot)
-        json_dict = {
-            "Id": replay.id,
-            "User": {
+        include_route = game.game_id == game_ids.GameIDs.TH08
+        if include_route:
+            route = replay.route
```
n-rook, line 20: This logic feels too complicated for me. Can't we just do

```
if replay.route:
  json_dict["Route"] = replay.route.GetName()
```
?

project/thscoreboard/replays/views/replay_list.py
```diff
@@ -48,6 +57,19 @@ def _get_filter_options_default(game: Game) -> dict[str, list[str]]:
     return {"Difficulty": all_difficulties, "Shot": all_shots}
 
 
+def _get_filter_options_th01(game: Game) -> dict[str, list[str]]:
```
n-rook, line 60: I won't block submission for this, but in the medium term it would be nice to figure out a way to share more code between all these _get_filter-options functions, since they look very similar right now.

n-rook (changes requested):

project/thscoreboard/replays/replays_to_json.py
jgdhs27: Unfortunately not as this would make `Route` the last column of the table

jgdhs27: Disregard my last comment. The issue is that IN Ex replays have no route and therefore end up misaligned. Although that's caused by the fact that we just insert the json data directly into the table in the frontend, which is some code we can/should improve

n-rook: Ah, I see. In that case, I guess the game_id check is fine (although don't we need it for hrtp too?). Still, though, I think we should just do

```
if condition:
  json_dict['Route'] = ...
```

rather than setting `include_route` up here and then using it below.

n-rook (approved):

PR #345, opened 2023-04-21
Use higher-resolution favicons.

n-rook:
I used a 48x48 image for the .ico file, and a 192x192 image for the .png. I used a bigger image for the .png because the .ico is only actually used on some pages, but it's sometimes fetched and then ignored. Plus, AFAIK .png is more efficient file format.

192x192 is way too big for any actual favicon, obviously, but it's nice because it divides evenly into 32 and 48. It's directly used in rare cases as well.

Visually, this looks much better than the old 16x16 version on my (150% DPI) monitor.


jgdhs27 (commented):
Before and After for me:
image
That's an improvement 👍

jgdhs27 (approved):

PR #348, opened 2023-04-23
Make games on leaderboard clickable

jgdhs27:
Make the game in the leaderboard located on the main page and the user page a clickable link.


n-rook (approved):
makes total sense, this just looks like an oversight. I'm surprised this works with no changes to replayTable.js

PR #349, opened 2023-04-23
Add medal emojis to top 3 scores

jgdhs27:
Add medal emojis (🥇, 🥈, 🥉) to the top 3 scores for a certain difficulty and shot.

image

In the future, I would like to add a player ranking by medal count, like royalflare had.

### DIFF
--- a/project/thscoreboard/replays/replays_to_json.py
+++ b/project/thscoreboard/replays/replays_to_json.py
@@ -1,16 +1,12 @@
-from collections import defaultdict
 from typing import Iterable
-from replays import models, game_ids
+from django.db.models import Window, F, Manager
+from django.db.models.functions import RowNumber
 from functools import lru_cache
+
+from replays import models, game_ids
 
 
 class ReplayToJsonConverter:
-    def __init__(self, include_medals=False) -> None:
-        self.include_medals = include_medals
-        self.replay_count_by_category: dict[tuple[int, models.Shot], int] = defaultdict(
-            int
-        )
-
     @lru_cache(maxsize=None)
     def _get_game(self, shot: models.Shot) -> models.Game:
         return shot.game
@@ -22,13 +18,7 @@ class ReplayToJsonConverter:
     def _convert_replay_to_dict(self, replay: models.Replay) -> dict:
         shot = replay.shot
         game = self._get_game(shot)
-        difficulty = replay.difficulty
-
-        score_prefix = ""
-        if self.include_medals:
-            self._update_replay_count_by_category(difficulty, shot)
-            rank = self.replay_count_by_category[(difficulty, shot)]
-            score_prefix = _get_medal_emoji(rank)
+        score_prefix = _get_medal_emoji(replay.rank)
 
         json_dict = {}
         json_dict["Id"] = replay.id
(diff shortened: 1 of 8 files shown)


project/thscoreboard/replays/replays_to_json.py
```diff
@@ -63,8 +77,21 @@ def _get_th17_additional_fields(self, shot: models.Shot) -> dict:
             "Goast": shot.GetSubshotName(),
         }
 
+    def _update_replay_count_by_category(self, difficulty: int, shot: models.Shot):
+        self.replay_count_by_category[(difficulty, shot)] += 1
+
     def convert_replays_to_serializable_list(
         self,
         replays: Iterable[models.Replay],
```
n-rook, line 85: seems like we've introduced an undocumented requirement that replays be ordered (in reverse order) by score, no?

n-rook (changes requested):

jgdhs27:
Ok, I came up with another approach which adds the ranking inside the db query. I find this cleaner, I hope it addresses your concerns.

jgdhs27:
Also, this makes it so that I could show the medals on the user page and front page without much difficulty, so I've done that.

project/thscoreboard/replays/replays_to_json.py
```diff
@@ -65,6 +69,28 @@ def _get_th17_additional_fields(self, shot: models.Shot) -> dict:
...
-        replays: Iterable[models.Replay],
+        ranked_replays: Iterable[models.Replay],
     ) -> list[dict[str, any]]:
-        return [self._convert_replay_to_dict(replay) for replay in replays]
+        return [self._convert_replay_to_dict(replay) for replay in ranked_replays]
+
+
+def add_rank_annotation_to_replays(
+    replays: Manager[models.Replay],
+) -> Manager[models.Replay]:
```
n-rook, line 79: Could you move this function over to models.py? We may want in places that aren't returning json, but it'd be strange to import "replays_to_json" in a place where we aren't converting replays to json. Plus, looking at the signature, I think you can define this as a method on a custom manager, if I remember Django right. Certainly calling replays.objects.some_query().annotate_with_rank() would be very convenient.

n-rook (changes requested):
Nice, this definitely looks like a good way to do it. Just one more comment

project/thscoreboard/replays/replays_to_json.py
jgdhs27: Done. Although I think the current implementation isnt perfect - I can only call `.annotate_with_rank()` on a `ReplayQuerySet` object but not a `ReplayManager` object, which feels wrong.

jgdhs27: disregard that, fixed that little issue too.

jgdhs27:
Ok, another revision done. `annotate_with_rank` is now part of the `ReplayQuerySet` object, leading to much nicer code. Added tests to that. I also fixed a bug where TASes and irregular replays were counted in the rankings.

n-rook (approved):
Nice!

I'm _slightly_ worried about performance, but it's probably fine :)

PR #352, opened 2023-04-25
Create "Claim" page that allow staff to assign royalflare replays to users

jgdhs27:
This is the first version of a tool to allow users to claim replays. For now, it is staff-only. In the future, I want to extend it so that users make their own requests.

image
image
image


project/thscoreboard/replays/templates/replays/claim_username.html
```diff
@@ -0,0 +1,15 @@
+{% extends "simple_centered.html" %}
+{% load i18n %}
+
+{% block centered-title %}<p>Claim a username</p>{% endblock %}
+
+{% block centered-content %}
+    <form method="post">
```
n-rook, line 7: What does this page look like if you get one of the usernames wrong? Does a good error appear?

project/thscoreboard/replays/templates/replays/claim_replays.html
```diff
@@ -0,0 +1,66 @@
...
+            <td>
+                {{replay.created | date:"Y-m-d"}}
+            </td>
+            <td>
+                {{replay.comment}}
+            </td>
+        </tr>
+    {% endfor %}
+    </table>
+    <input type="submit" value="Confirm" style="margin-top: 8px;">
```
n-rook, line 63: Even though there are staff-only pages, please still rely on CSS rather than the style= rule.

project/thscoreboard/users/views/claim.py
```diff
@@ -0,0 +1,89 @@
...
+    if request.method == "GET":
+        return _render_claim_username_form(request)
+    elif (
+        request.method == "POST" and request.POST.get("royalflare_username") is not None
+    ):
+        return _render_claim_replay_form(request)
+    else:
+        selected_replays = _get_selected_replays(request)
+        user = _get_user(request.POST["silentselene_username"])
+        _asign_selected_replays_to_user(selected_replays, user)
```
n-rook, line 26: *assign

project/thscoreboard/users/views/claim.py
```diff
@@ -0,0 +1,89 @@
...
+        request,
+        "replays/claim_username.html",
+        {
+            "form": form,
+        },
+    )
+
+
+def _get_selected_replays(request: WSGIRequest) -> list[models.Replay]:
+    selected_replays = []
```
n-rook, line 42: Since you're already defining Django form objects, why not lean on them for parsing the requests? When I see something like this, I worry that something weird could sneak into the request object dictionary and cause unusual behavior.

project/thscoreboard/users/views/claim.py
```diff
@@ -0,0 +1,89 @@
...
+        if replay_was_selected:
+            replay = models.Replay.objects.get(id=replay_id)
+            selected_replays.append(replay)
+    return selected_replays
+
+
+def _asign_selected_replays_to_user(
+    replays: Iterable[models.Replay], user: User
+) -> None:
+    for replay in replays:
```
n-rook, line 54: What if the replay already belongs to someone else? I think we should wrap this in a transaction (or one transaction per replay, maybe) and check to make sure we aren't overriding an existing owner.

n-rook (changes requested):
Looks good, just a few comments

project/thscoreboard/users/views/claim.py
jgdhs27: fixed

project/thscoreboard/replays/templates/replays/claim_username.html
jgdhs27: Wrong ss username: you get brought to one of those error pages with the title
DoesNotExist at /users/claim
User matching query does not exist.

Wrong rf username: No error, instead there are simply no replays to select.

What does a good error mean? I have no idea what the right thing to do here is.

project/thscoreboard/users/views/claim.py
jgdhs27: The query to gat all replays already filters things, although you're right that this could cause bugs and wrapping it in a transaction is safer. Done. Btw, I'm again not sure what to do in case of an error.

project/thscoreboard/replays/templates/replays/claim_replays.html
jgdhs27: done

project/thscoreboard/replays/templates/replays/claim_username.html
n-rook: I was being vague in case you did things this way intentionally and I was wrong about what it'd look like, but I do actually have an opinion here: You should use the Django form API's error features, which provide a convenient way to attach errors to form fields in Python and have them display nicely on the website. See https://docs.djangoproject.com/en/4.2/topics/forms for details, especially [this section](https://docs.djangoproject.com/en/4.2/topics/forms/#rendering-form-error-messages). This is what we do on most of the site; for example see replays/views/create_replay.py for some of these.

n-rook (commented):
I think you didn't upload your latest commit :)

project/thscoreboard/users/views/claim.py
jgdhs27: Took me a while to understand how to do this, but its done :)

project/thscoreboard/replays/templates/replays/claim_username.html
jgdhs27: added helpful error messages to the form

n-rook (approved):
Nice, looks good! It took me awhile to get comfortable with Django forms too.

PR #354, opened 2023-04-26
Update translation files.

n-rook:
There was nothing interesting in these updates, so I doubt this will have any noticeable impact.


jgdhs27 (approved):
didnt really review it but i trust

PR #360, opened 2023-04-28
Fix user page rankings

### LINKED ISSUE #353: Medals on user page are broken
It looks like the replays are only ranked against other replays on the user page. Not sure why.

jgdhs27:
Currently, we call `.annotate_with_rank().filter(user=user)` to try to get user replays ranked. This doesn't work though, as the rank will apply to the filtered replays. Not really sure why, I don't understand DB stuff.

As a workaround, store ranks for all replays in a python dict, then apply them to a filtered queryset. This is cringe. Performance dropped by about 200ms. I'm sure there's a better way but idk what it is.

Fixes #353

### DIFF
--- a/project/thscoreboard/replays/test_rank_replays.py
+++ b/project/thscoreboard/replays/test_rank_replays.py
@@ -1,0 +1,54 @@
+from unittest.mock import patch
+
+from replays.rank_replays import add_global_rank_annotations
+from replays.testing import test_case
+from replays import models
+from replays.testing import test_replays
+
+
+class GlobalRankAnnotationsTestCase(test_case.ReplayTestCase):
+    def setUp(self):
+        super().setUp()
+        self.user = self.createUser("some-user")
+
+    def _create_n_replays(self, n: int, is_tas: bool = False) -> None:
+        for i in range(n):
+            with patch(
+                "replays.constant_helpers.CalculateReplayFileHash"
+            ) as mocked_hash:
+                mocked_hash.return_value = bytes(i)
+                test_replays.CreateAsPublishedReplay(
+                    "th10_normal",
+                    user=self.user,
+                    score=1_000_000_000 - 100_000_000 * i,
+                    difficulty=0,
+                    category=models.Category.TAS
+                    if is_tas
+                    else models.Category.STANDARD,
+                )
+
+    def testRanksAgainstAllReplays(self):
+        self._create_n_replays(3)
+        replay_subset = models.Replay.objects.order_by("-score")[1:]
+        add_global_rank_annotations(replay_subset)
+
+        self.assertEqual(replay_subset[0].rank, 2)
+        self.assertEqual(replay_subset[1].rank, 3)
+
+    def testGivesNegativeRankToReplaysOutsideTopThree(self):
+        self._create_n_replays(5)
+        replay_subset = models.Replay.objects.order_by("-score")[3:]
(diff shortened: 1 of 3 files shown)


n-rook:
This is too slow :/ Especially since if it's taking 200ms now, I bet the big-O answer will be even worse once we have more replays in the database. I'd rather see us just stop displaying ranks on pages other than the full scoreboard for now.

In the long term this problem does seem fixable; it should be possible to get the ranks of replays in a reasonable amount of time. Especially since we can clamp the big-O from both ends; we aren't fetching that many rows here, and in all cases we don't care about the ranks of anything but the top replays. But it might be tricky. These pages might be helpful in figuring out how to do it:

https://docs.djangoproject.com/en/4.2/topics/db/aggregation/#combining-multiple-aggregations
https://stackoverflow.com/questions/56567841/django-count-and-sum-annotations-interfere-with-each-other

Remember too that you can uncomment part of settings.py to log all SQL queries, which should make it clearer what's really going wrong.

jgdhs27:
Thanks for the links. I will try again some time next week, but if I can't get it to work, I'll just remove the feature for now.

n-rook:
You can always disable it for now and come back later when you have time to figure it out, too

jgdhs27:
Is using Django 4.2 an option? https://stackoverflow.com/questions/68289137/filtering-on-a-window-function-in-django

n-rook:
I don't see why not, if it works

n-rook:
However definitely do not use Django 4.2 in this PR; if it works, send that change alone as a separate PR, and then merge it in here

jgdhs27:
I still haven't found a way to get this to work with just a DB query. I'm still using a python dict, but I tried to get better performance by adding only top-3 replays to the dict.

n-rook (commented):
How fast is it now? I didn't import rf on my personal machine, so I don't have enough data to tell when I tried it myself.

jgdhs27:
On my machine, this function went from 0.045s to 0.145s.

n-rook:
Okay, I think that's fine then. In reality doing it this way isn't really that much slower than doing it on the db side; the main disadvantage comparatively is that it's kinda hacky (since you'd have to redo the same logic on other pages)

n-rook (changes requested):
This logic is complicated enough that I want a unit test for it. Could you extract the new code into a public function in some other module (or a new module), and then write a unit test for it? That'd make it easier to use it elsewhere too.

jgdhs27:
Ok, I extracted it and added some tests

project/thscoreboard/replays/test_rank_replays.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class GlobalRankAnnotationsTestCase(test_case.ReplayTestCase):
+    def setUp(self):
+        super().setUp()
+        self.user = self.createUser("some-user")
+
+    def _create_n_replays(self, n: int, is_tas: bool = False) -> None:
+        for i in range(n):
+            with patch(
```
n-rook, line 16: nit: Rather than patching n times, I think it'd be simpler and faster to patch once and just set return_value in each loop iteration.

n-rook (approved):
Great!

project/thscoreboard/replays/test_rank_replays.py
jgdhs27: oh so swap the order of the `for` and `with`? sure.

PR #365, opened 2023-04-29
Don't use pytz in import_royalflare.

n-rook:
I think our tests would have caught this but sadly we don't have them!


jgdhs27 (approved):
Sorry!

n-rook:
npnp, blame pip :)

PR #368, opened 2023-04-29
Rename the "Regular" enum for both replay type and category.

### LINKED ISSUE #216: Rename "regular" replay type or category
Currently, both replay type and category have an option called "Regular", which is insanely confusing. We should rename one of them.

Actually, I don't like "regular" for either name, now that I think about it, so I'd be down to rename both.

The "Regular" replay type is in contrast to stage practice, spell practice, and so on. For this one, we should pick a name that makes it obvious what it is. Maybe "Full Run" or "Full Game"?

The category is in contrast to "Tool-Assisted". For this one, maybe "Standard"? As in, "played according to standards".

n-rook:
Previously, the "Regular" category indicated that a run was played by a player with no special tools (in contrast to something like a TAS run). The "Regular" category in replay types indicated a run was a full game run.

The "Regular" category has been renamed to "Standard". The replay type has been renamed to "Full Game".

If anyone has better names, let me know.

Note that the migration here is basically fake... for some reason Django generates a migration but it doesn't actually do anything (because the enum's actual value hasn't changed, just the string representation).

Fixes #216


project/thscoreboard/replays/models.py
```diff
@@ -124,8 +124,8 @@ class Category(models.IntegerChoices):
 class ReplayType(models.IntegerChoices):
     """Type of replay (regular, spell practice, etc)"""
```
jgdhs27, line 125: update comment

jgdhs27 (approved):
Naming choices make sense to me

Issue #370, opened 2023-04-30
Bad performance when loading a large amount of replays


jgdhs27:
Performance on https://www.silentselene.net/replays/th08 is bad. Loading the json takes about 1.7s. On my local setup, the request takes just 0.35s.

One possible improvement would be to stream the json data, or at lest load it in batches.
As a starting point: some time ago I tried modifying the API calls to first load just the first 50 files, and then load the remaining files. The code for that is on [this branch](https://github.com/jgdhs27/thscoreboard/tree/api-speedup). This worked but had no measurable performance impact. I can imagine it'll help on a non-local setup.

It would also help to request replays immediately (see issue #351)

jgdhs27:
Also consider reducing the amount of data sent. This could be done by compressing, or by removing unnecessary data. Maybe the data should be in csv format or similar.

n-rook:
I think gzip is probably the first thing to try. json tends to be very compressible. I suspect this will have a significant benefit.

jgdhs27:
With #373 and #390 landed I consider this fixed.

PR #373, opened 2023-05-02
GZ compress replay response

### LINKED ISSUE #372: Filters are not applied when replays are first loaded
To replicate, go on https://www.silentselene.net/replays/th08 and use any filter before the page loads.

Should be trivial to fix

jgdhs27:
Use GZ to compress the json sent by the server. This reduces the amount of data sent by about 84%, or from 2.2MB to 350kB for IN.

Related: #370
Fixes #372

### DIFF
--- a/project/thscoreboard/replays/views/replay_list.py
+++ b/project/thscoreboard/replays/views/replay_list.py
@@ -1,11 +1,8 @@
 """Contains views which list various replays."""
 
-import gzip
-import json
-from django.http import HttpResponse
+from django.http import JsonResponse
 from django.views.decorators import http as http_decorators
 from django.shortcuts import get_object_or_404, render
-
 
 from replays import models
 from replays import game_ids
@@ -15,13 +12,11 @@ from replays.replays_to_json import ReplayToJsonConverter
 
 @http_decorators.require_safe
 def game_scoreboard_json(request, game_id: str):
-    replays = _get_all_replay_for_game(game_id)
     converter = ReplayToJsonConverter()
-    replays_json = json.dumps(converter.convert_replays_to_serializable_list(replays))
-    replays_json_compressed = gzip.compress(replays_json.encode("utf-8"), 5)
-    response = HttpResponse(replays_json_compressed)
-    response["Content-Encoding"] = "gzip"
-    return response
+    replays = _get_all_replay_for_game(game_id)
+    return JsonResponse(
+        converter.convert_replays_to_serializable_list(replays), safe=False
+    )
 
 
 @http_decorators.require_safe
--- a/project/thscoreboard/replays/views/replay_list.py
+++ b/project/thscoreboard/replays/views/replay_list.py
@@ -1,6 +1,8 @@
 """Contains views which list various replays."""
 
-from django.http import JsonResponse
+import gzip
+import json
+from django.http import HttpResponse
 from django.views.decorators import http as http_decorators
 from django.shortcuts import get_object_or_404, render
 
@@ -13,12 +15,13 @@ from replays.replays_to_json import ReplayToJsonConverter
 
 @http_decorators.require_safe
 def game_scoreboard_json(request, game_id: str):
-    converter = ReplayToJsonConverter()
     replays = _get_all_replay_for_game(game_id)
-    return JsonResponse(
-        converter.convert_replays_to_serializable_list(replays),
-        safe=False,
-    )
+    converter = ReplayToJsonConverter()
+    replays_json = json.dumps(converter.convert_replays_to_serializable_list(replays))
+    replays_json_compressed = gzip.compress(replays_json.encode("utf-8"), 5)
+    response = HttpResponse(replays_json_compressed)
+    response["Content-Encoding"] = "gzip"
+    return response
 
 
 @http_decorators.require_safe
--- a/project/thscoreboard/shared_content/static/js/replays/replayTable.js
+++ b/project/thscoreboard/shared_content/static/js/replays/replayTable.js
@@ -12,29 +12,27 @@ var allReplays = [];
 
 requestAndInitializeReplays();
 
-function requestAndInitializeReplays() {
+async function requestAndInitializeReplays() {
   const requestUri = window.location.pathname === '/'
     ? window.location.origin + '/replays/index/json'
     : window.location.origin + window.location.pathname + "/json";
 
-  fetch(requestUri, {
-    headers: {
-      'Accept-Encoding': 'gzip'
-    },
-    priority: 'high'
-  })
-  .then(response => {
+  try {
+    const response = await fetch(requestUri, {
+      priority: 'high'
+    });
+
     if (!response.ok) {
       replayTableBodyHtml.innerHTML =
         '<p style="color: red;">Failed to load replays</p>';
+      return;
     }
-    return response.json();
-  })
-  .then(replayJson => {
-    allReplays = replayJson
+
+    const replayJson = await response.json();
+    allReplays = replayJson;
     const filteredReplays = filterReplays(activeFilters, allReplays);
     constructAndRenderReplayTable(filteredReplays);
-  }).catch(function() {});
+  } catch (error) {}
 }
 
 function onClick(elm) {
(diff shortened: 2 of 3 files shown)


jgdhs27:
image
based on this chart, i think compression could reduce the time for "Comtent Download" by about 250ms.

I noticed also that the request API lets me choose the priority, and I set its priority to high. (Done in the second commit.) Considering that the connection was stalled for 540ms, I think there is a lot more room for improvement right there!

project/thscoreboard/shared_content/static/js/replays/replayTable.js
```diff
@@ -10,26 +10,31 @@ const displayedColumns = [
...
+  fetch(requestUri, {
+    headers: {
+      'Accept-Encoding': 'gzip'
+    },
+    priority: 'high'
+  })
+  .then(response => {
+    if (!response.ok) {
+      replayTableBodyHtml.innerHTML =
+        '<p style="color: red;">Failed to load replays</p>';
```
n-rook, line 29: If you keep putting style tags on HTML, I'm going to keep telling you to use CSS instead :p

In this case, I think the best thing to do would actually be to put this error in the real html template, hide it by default, and unhide it in JS when appropriate (by removing a class or something like that)

project/thscoreboard/shared_content/static/js/replays/replayTable.js
```diff
@@ -10,26 +10,31 @@ const displayedColumns = [
...
-    } else {
-      document.getElementById('replay-table').innerHTML = '<p style="color: red;">Failed to load replays</p>';
+requestAndInitializeReplays();
+
+function requestAndInitializeReplays() {
+  const requestUri = window.location.pathname === '/'
+    ? window.location.origin + '/replays/index/json'
+    : window.location.origin + window.location.pathname + "/json";
+
+  fetch(requestUri, {
```
n-rook, line 20: might be easier to call this as an async function and use await

project/thscoreboard/shared_content/static/js/replays/replayTable.js
```diff
@@ -10,26 +10,31 @@ const displayedColumns = [
...
+    headers: {
+      'Accept-Encoding': 'gzip'
```
n-rook, line 22: don't think you should do this (or have to do this)--- browser will set Accept-Encoding for you

project/thscoreboard/replays/views/replay_list.py
```diff
@@ -13,12 +15,13 @@
...
 def game_scoreboard_json(request, game_id: str):
-    converter = ReplayToJsonConverter()
     replays = _get_all_replay_for_game(game_id)
-    return JsonResponse(
-        converter.convert_replays_to_serializable_list(replays),
-        safe=False,
-    )
+    converter = ReplayToJsonConverter()
+    replays_json = json.dumps(converter.convert_replays_to_serializable_list(replays))
+    replays_json_compressed = gzip.compress(replays_json.encode("utf-8"), 5)
```
n-rook, line 21: I don't think you should need to call compress() yourself--- we should use middleware for this. That way we can respond appropriately to the Accept-Encoding header of the request.

n-rook (changes requested):

project/thscoreboard/thscoreboard/settings.py
```diff
@@ -265,3 +265,10 @@ def _GetAllowedHosts():
 
 DISCORD_WEBHOOK_ID = os.environ.get("DISCORD_WEBHOOK_ID", None)
 DISCORD_WEBHOOK_TOKEN = os.environ.get("DISCORD_WEBHOOK_TOKEN", None)
+
+MIDDLEWARE = [
```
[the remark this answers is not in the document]
jgdhs27: oops, fixed

project/thscoreboard/shared_content/static/js/replays/replayTable.js
jgdhs27: done

project/thscoreboard/shared_content/static/js/replays/replayTable.js
jgdhs27: you're right, removed it

project/thscoreboard/replays/views/replay_list.py
n-rook: I tried this, and I think GZipMiddleware should just do this automatically on its own. You might have to put it at the top of the middleware section (https://docs.djangoproject.com/en/4.2/ref/middleware/#module-django.middleware.gzip).

It's better to do this in middleware because it separates the details of how the information is transmitted from the information itself. If I go look at game_scoreboard_json, I don't want to have to read about how it's encoded--- that can be handled elsewhere. Plus, it allows us to take advantage of the existing implementation to do things like not gzip if the client doesn't accept gzip encoding.

project/thscoreboard/shared_content/static/js/replays/replayTable.js
n-rook: Oops! Well, I still don't like it :p but you're right that it's not relevant to this PR.

project/thscoreboard/replays/views/replay_list.py
jgdhs27: I have no idea how to accomplish that

n-rook: no I mean, I'm pretty sure if you straight up just delete all the gzip code, the gzip middleware that's already installed will just gzip it for you anyway

jgdhs27: still can't figure it out

n-rook: What are you having trouble with?

Middleware in django (https://docs.djangoproject.com/en/4.2/topics/http/middleware/) basically consists of logic that operates both before and after each view. that's what makes it a good fit for something like compression; we want to compress everything (or at least, we want to compress a lot of things, if they're big enough to be worthwhile), so putting the gzipmiddleware at the top of the middleware list in settings.py allows that to happen automatically for lots of requests, which is better than doing it for only one request. If you do that, even without the code here in replay_list, gzipmiddleware should automatically compress this request for you, and you should be able to see that in something like the Chrome network inspector (because you can look for the gzip tag on the response).

Let me know (or feel free to ping me) if there's anything else you want more information about here from me.

jgdhs27: For instance if I remove the compression code in python,
```python
    replays_json = json.dumps(converter.convert_replays_to_serializable_list(replays))
    # replays_json_compressed = gzip.compress(replays_json.encode("utf-8"), 5)
    response = HttpResponse(replays_json)
    response["Content-Encoding"] = "gzip"
    return response
```
it doesn't work. I don't know what I'm supposed to do.

n-rook: Oh, I see the issue. You aren't supposed to set Content-Encoding yourself either. If you do, the GZipMiddleware assumes that you already encoded it, so it doesn't do anything; it only works on stuff returned without a Content-Encoding. Instead you should literally just revert replay_list.py and not change it. The main benefit of all this middleware stuff is that you don't have to change the function if you want to add higher-level functionality.

You can actually go look at the source code for Django yourself if you don't understand what it's doing:

https://github.com/django/django/blob/4bf4222010fd8e413963c6c873e4088614332ef9/django/middleware/gzip.py

When I was trying to figure out this middleware, I went into my local installed Django and put stuff like `print('hi im gzip')` in the source so that I could be sure it was actually working.

I uploaded what I did at https://github.com/n-rook/thscoreboard/tree/nrook/gzip-middleware-only if you want to take a look there too.

32th-System: Wouldn't that gzip compress everything? What if we don't want every page to be gzip compressed?

n-rook: Well, why wouldn't you?

In theory even if we didn't want to gzip everything, I would want to do that by writing our own version of GZipMiddleware and deciding whether or not to gzip stuff based on the responses. I definitely want all _big_ things to be gzipped by default. (or we could install middleware to do brotli or something instead)

jgdhs27:
Added nrook's in which everything gzip-related is handled using middleware

n-rook (approved):
Well, of course I approve :p

PR #377, opened 2023-05-04
Test import royalflare and fix creation timestamp bug

jgdhs27:
Add tests for import_royalflare.py.

While doing so, I found a bug: I mixed up the uploaded timestamp with the replay timestamp. This is fixed.

### DIFF
--- a/project/thscoreboard/replays/test_import_royalflare.py
+++ b/project/thscoreboard/replays/test_import_royalflare.py
@@ -1,0 +1,93 @@
+from datetime import datetime, timezone
+from pathlib import Path
+import unittest
+
+from replays.management.commands.import_royalflare import (
+    import_royalflare,
+    parse_timestamp_from_json,
+    parse_replay_path_from_json,
+)
+from replays.testing import test_case, test_replays
+from replays import models
+
+
+class ParseTimestampTest(unittest.TestCase):
+    def testParseDateOnly(self):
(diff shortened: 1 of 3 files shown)


project/thscoreboard/replays/test_import_royalflare.py
```diff
@@ -0,0 +1,107 @@
...
+            "player": "Test Player",
+            "comment": "Test Comment",
+            "replay": "/replays/path/to/replay.rpy",
+            "uploaded": "2022/01/01",
+        }
+
+        with patch(
+            "replays.management.commands.import_royalflare.parse_replay_path_from_json"
+        ) as mock_parse_replay_path_from_json:
+            test_replay_path = test_replays.TEST_REPLAY_LOCATION / "th10_normal.rpy"
```
n-rook, line 57: wait, can't you just pass TEST_REPLAY_LOCATION in as the value of the second parameter to import_royalflare? that seems much simpler than using mock.patch()

n-rook (approved):
Nice, thanks!

(Me realizing the bugfix means I should delete and reimport all of the replays: "noooooooooooooo")

project/thscoreboard/replays/test_import_royalflare.py
jgdhs27: ah, you're right, that makes things simpler

PR #389, opened 2023-05-07
Add a language switcher to the header.

n-rook:
image

I think this is fundamentally the proper backend approach, but it doesn't look great on the frontend. But I wanted to get in a reasonable switcher quickly so we can actually test the Japanese translations.


jgdhs27 (approved):
Can't try it locally but I trust that it works, I'll just check it on staging lol

jgdhs27:
Looks good!

32th-System:
it does work

PR #390, opened 2023-05-09
Stream replays from DB to frontend

jgdhs27:
Stream the replays from the database query all the way to displaying it in the user's browser. This should improve performance, especially when a lot of replays are requested.

It's hard to measure the impact of this change, because performance is fine locally. I tried testing the code using every replay in my DB (50,000 total), but the numbers don't mean much:
Before: Replays appear after 1747ms
After: Replays appear after 923ms
I would like to push this to staging with javascript logging so that we can better understand what it does and see its impact.

### DIFF
--- a/project/thscoreboard/replays/replays_to_json.py
+++ b/project/thscoreboard/replays/replays_to_json.py
@@ -1,6 +1,8 @@
 import json
-from typing import Iterable, Iterator
+from typing import Iterable
 from functools import lru_cache
+
+from django.http import StreamingHttpResponse
 
 from replays import models, game_ids
 
@@ -69,13 +71,29 @@ class ReplayToJsonConverter:
             "Goast": shot.GetSubshotName(),
         }
 
-    def convert_replays_to_serializable_list(
-        self,
-        ranked_replays: Iterable[models.Replay],
-    ) -> Iterator[dict[str, any]]:
-        for replay in ranked_replays:
-            json_string = json.dumps(self._convert_replay_to_dict(replay))
-            yield f"{json_string}\n".encode("utf-8")
+    def convert_replay_to_json_string(self, replay: models.Replay) -> str:
+        replay_dict = self._convert_replay_to_dict(replay)
+        return f"{json.dumps(replay_dict)}\n".encode("utf-8")
+
+
+def convert_replays_to_json_strings(
+    ranked_replays: Iterable[models.Replay],
+) -> Iterable[str]:
+    converter = ReplayToJsonConverter()
+    return (
+        converter.convert_replay_to_json_string(replay) for replay in ranked_replays
+    )
+
+
+def stream_json_strings_to_http_reponse(
+    replay_strings: Iterable[str],
+) -> StreamingHttpResponse:
+    response = StreamingHttpResponse(
+        iter(replay_strings),
+        content_type="application/json",
+    )
+    response["Content-Disposition"] = 'attachment; filename="output.json"'
+    return response
 
 
 def _get_medal_emoji(rank: int) -> str:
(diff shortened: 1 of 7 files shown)


project/thscoreboard/replays/test_replays_to_json.py
```diff
@@ -66,9 +66,9 @@ def testConvertReplaysToJsonString(self):
 
         with test_utilities.OverrideTranslations():
             converter = ReplayToJsonConverter()
-            json_data = converter.convert_replays_to_serializable_list(replays)
-
-        assert len(json_data) == 2
+            json_data = [
+                converter._convert_replay_to_dict(replay) for replay in replays
```
n-rook, line 70: In general, it's best to avoid testing private methods; could you test this instead by calling convert_replays_to_json_strings?

project/thscoreboard/replays/replays_to_json.py
```diff
@@ -68,11 +71,29 @@ def _get_th17_additional_fields(self, shot: models.Shot) -> dict:
...
+def convert_replays_to_json_strings(
+    ranked_replays: Iterable[models.Replay],
+) -> Iterable[str]:
+    converter = ReplayToJsonConverter()
+    return (
+        converter.convert_replay_to_json_string(replay) for replay in ranked_replays
+    )
+
+
+def stream_json_strings_to_http_reponse(
```
n-rook, line 88: Could you move this function to the view? I think it's cleanest if a module like this one doesn't actually handle HTTP concepts directly, but rather just returns data objects.

n-rook (changes requested):
Sorry for the review delay, busy week.

I don't really have strong feelings about whether to do this, but if you want, you can keep the old function too, and use an environment variable to call staging based on whether we are in staging or not. that is, just add something to settings.py called IS_STAGING or something, set it to os.environ.get('IS_STAGING', False), and call a different method depending on the situation. this would make the js more complicated though, and plus you'd have to figure out a way to actually get the env variable to the js.

project/thscoreboard/replays/test_replays_to_json.py
32th-System: You could also make the method public

jgdhs27:
> Sorry for the review delay, busy week.
> 
> I don't really have strong feelings about whether to do this, but if you want, you can keep the old function too, and use an environment variable to call staging based on whether we are in staging or not. that is, just add something to settings.py called IS_STAGING or something, set it to os.environ.get('IS_STAGING', False), and call a different method depending on the situation. this would make the js more complicated though, and plus you'd have to figure out a way to actually get the env variable to the js.

imo, it feels like it would be quite a bit of work, I really think its easier to be a little careful with what we push. It's not like it would be a disaster if this is pushed to production, there are only a few unnwanted console.log's.

project/thscoreboard/replays/replays_to_json.py
```diff
@@ -68,11 +71,29 @@ def _get_th17_additional_fields(self, shot: models.Shot) -> dict:
             "Goast": shot.GetSubshotName(),
         }
 
-    def convert_replays_to_serializable_list(
-        self,
-        ranked_replays: Iterable[models.Replay],
-    ) -> list[dict[str, any]]:
-        return [self._convert_replay_to_dict(replay) for replay in ranked_replays]
+    def convert_replay_to_json_string(self, replay: models.Replay) -> str:
```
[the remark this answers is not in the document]
jgdhs27: good spot

project/thscoreboard/replays/replays_to_json.py
n-rook: Oh, you're right. I actually kinda feel like maybe this should be one endpoint eventually, and we should customize its behavior from the frontend side. Not in this CL though.

n-rook (approved):

PR #391, opened 2023-05-09
Allow users to claim replays

jgdhs27:
Allow non-staff users to claim replays. When a non-staff user visits /users/claim, they can choose a royalflare username, and then choose some replays to claim. They have to provide contact details alongside their request. These claims are currently just written to the database and are not yet processed in any way.


project/thscoreboard/users/models.py
```diff
@@ -599,3 +600,36 @@ def CleanUp(cls, now: datetime.datetime) -> None:
...
+
+
+class ClaimReplayRequestManager(models.Manager):
+    def get_active_requests_for_user(self, user):
+        return self.filter(user=user, request_status=RequestStatus.SUBMITTED)
+
+    def create(self, *args, **kwargs):
+        user = kwargs.get("user", None)
+        if user and self.get_active_requests_for_user(user).count() >= 5:
+            raise ValidationError("Each user may only have 5 active requests at once.")
```
n-rook, line 616: Is this a good idea? I would have expected some people would really want to claim a ton of replays, if they had a bunch on royalflare already.

n-rook (approved):
Looks good to me

project/thscoreboard/users/models.py
jgdhs27: It's not a limit on the number of replays requested, but a limit on the number of requests. Each request can have many replays (but only one username).

PR #392, opened 2023-05-14
Allow users to list/unlist replays

jgdhs27:
Adds a new flag for each replay, which can be set by the user, to list or unlist replays. unlisted replays are hidden on the main page and the game leaderboards, but still shown on the user profile.

The motivation is that we will soon allow users to claim replays that were imported from royalflare. They might want to hide those replays if they have beaten their pbs. Then unlisting is nicer than deleting, since they can still access their replays, and they are still kept in our database.


32th-System:
we already got a "Category" field but the problem with it is that it makes "Tool-Assisted" and "Unlisted" mutually exclusive. Should we get rid of it?

jgdhs27:
> we already got a "Category" field but the problem with it is that it makes "Tool-Assisted" and "Unlisted" mutually exclusive. Should we get rid of it?

I think the distinction is still good to have. I want to add a wr history page that includes unlisted but not TAS/Unusual replays.

project/thscoreboard/replays/constant_helpers.py
```diff
@@ -44,9 +44,9 @@ def GetModelInstancesForReplay(
     return ReplayConstantModels(game=shot.game, shot=shot, route=route)
 
 
-def CheckReplayFileDuplicate(file):
+def GetReplayFileWithSameHash(file) -> Optional[models.ReplayFile]:
```
n-rook, line 47: This is two logical PRs! The first PR implements a feature to allow users to hide replays from the scoreboard if they don't want them displayed. The second improves the error shown if a duplicate replay is open. Both are good ideas, but I am going to insist you split them into separate PRs.

project/thscoreboard/replays/templates/replays/replay_details.html
```diff
@@ -120,6 +120,17 @@ <h2><a href="/replays/{{ game_id}}">{{ game_name }}</a></h2>
         <a href="{{replay.video_link}}">Watch replay on external site</a>
     </div>
     {% endif %}
+    {% if can_edit %}
+        {% if is_listed %}
+        <div>
+            <a href="/replays/{{ game_id }}/{{ replay.id }}/unlist">Hide replay on the leaderboards</a>
```
n-rook, line 126: We used a separate page for Delete because we didn't want people doing it by accident, but I'm not really convinced a separate page is necessary here. It's not a big deal if someone accidentally unlists a replay, because they can just hit the button again. As such, I think we are better off directly using a button here that immediately does it.

n-rook (changes requested):
From a product perspective, I do think this is a good idea. As morth noted, the category field represents what a replay _is_: it's an inherent property of the replay. It would be wrong for someone to upload a replay under the wrong category. But the field this PR adds doesn't really have a "right" value; it's just based on how the player feels.

project/thscoreboard/replays/constant_helpers.py
jgdhs27: Ok, this change now exists on #398

project/thscoreboard/replays/templates/replays/replay_details.html
jgdhs27: Done. Now it will simply reload the page.

project/thscoreboard/replays/views/view_replay.py
```diff
@@ -158,6 +159,32 @@ def delete_replay(request, game_id: str, replay_id: int):
     )
 
 
+@http_decorators.require_http_methods(["GET", "HEAD", "POST"])
+@auth_decorators.login_required
+def list_replay(request, game_id: str, replay_id: int):
+    _set_listed(request, game_id, replay_id, True)
+    return replay_details(request, game_id, replay_id)
```
n-rook, line 166: If I were doing this I would probably 304 over to replay_details instead, but thinking about it, I don't think there's anything wrong with this way either.

n-rook (approved):

project/thscoreboard/replays/views/view_replay.py
jgdhs27: Didn't know about that! but I tried it out and it didnt seem to update

PR #398, opened 2023-05-15
Show a more helpful error on duplicate replay upload

jgdhs27:
I want to add unlisted replays. But then, a replay could be rejected for being a duplicate of a replay not shown on the leaderboards. This might cause some confusion. Therefore, we should at least tell the uploader where the duplicate replay is.

The new error message is: "This replay already exists. /replays/th08/18"
Consider the message temporary. I didn't want to write any new html, but felt that it was necessary to display the info in some form.


n-rook (approved):
Nice!

PR #400, opened 2023-05-16
Review replay claims

jgdhs27:
Allow users to review claims that are in the database. Users are able to delete claims, staff are additionally able to approve them. If a claim is approved, the replays are assigned.

This PR does not notify anyone about reviews or link to them in any way.


jgdhs27:
oh boy, 2 different test runs give 2 different failures
https://github.com/n-rook/thscoreboard/actions/runs/4996328526/jobs/8949591491#step:7:24
https://github.com/n-rook/thscoreboard/actions/runs/4996328526/jobs/8949370095#step:7:24

jgdhs27:
> oh boy, 2 different test runs give 2 different failures https://github.com/n-rook/thscoreboard/actions/runs/4996328526/jobs/8949591491#step:7:24 https://github.com/n-rook/thscoreboard/actions/runs/4996328526/jobs/8949370095#step:7:24 (it works on my machine)

The tests where I load the page as a staff don't seem to work on github CI. I might just remove them.

project/thscoreboard/users/views/claim.py
```diff
@@ -38,6 +39,53 @@ def claim(request: WSGIRequest) -> HttpResponse:
...
+        ).first()
+    )
+    if not _check_user_can_see_claim(request.user, claim):
+        raise PermissionDenied
+
+    if request.method == "POST":
+        form = forms.ClaimReplaysForm(
+            request.POST, replays=replay_models.Replay.objects.all()
+        )
+        if form.is_valid():
```
n-rook, line 56: If the form isn't valid, I believe we return None from this function, no? That seems not ideal.

project/thscoreboard/users/forms.py
```diff
@@ -193,6 +193,7 @@ class ClaimReplaysForm(forms.Form):
         label=_("contact_info"), required=True, max_length=200
     )
     silentselene_username = forms.CharField(widget=forms.HiddenInput(), required=False)
+    submit = forms.CharField(widget=forms.HiddenInput(), required=False)
```
n-rook, line 196: This pattern is a bit unusual and surprising. As such, please document it, and use a ChoiceField or something so that the only legal values for it are the ones we expect.

n-rook (changes requested):
Thanks for doing this. Sorry it took me awhile to look at it; I've been quite busy.

n-rook (changes requested):
Thanks for doing this. Sorry it took me awhile to look at it; I've been quite busy.

project/thscoreboard/users/forms.py
jgdhs27: made it into a ChoiceField with only 3 valid choices

project/thscoreboard/users/views/claim.py
n-rook: This is one of the good things about Django's form API! If you load a form from a POST, and it isn't valid, the thing to do is to pass that form instance into render() with the same template we used for initially rendering the form. That way, the user gets the same form, except with inline errors reflecting what went wrong. There are some existing instances of doing this in the codebase, like in create_replay.

n-rook (approved):

PR #401, opened 2023-05-18
Allow users to unclaim replays

jgdhs27:
Adds a new button to unclaim replays. This button is visible only if:
1. The replay is imported
1. The replay is claimed
1. The user owns the replay or is staff

After clicking it, you are redirected to a confirmation page:
image
Confirming executes the action.


n-rook (approved):
fwiw I don't think we need to let staff do this

PR #402, opened 2023-05-18
Send a discord message when a user creates a new claim replay request

jgdhs27:
Sends a message like this
"User morth1 has requested ownership of some replays. You can review this request at http://localhost:8000/users/claim/17 "
to discord when a new claim is created.

The discord token secret and discord channel id are stored as environment variables.


jgdhs27:
only the last two commits are important

32th-System:
This could be done with a webhook. We're already using one for the deployment announcements

jgdhs27:
> This could be done with a webhook. We're already using one for the deployment announcements

Oh? I'll look into it

PR #403, opened 2023-05-18
(Webhook) Send a discord message when a user creates a new claim replay request

jgdhs27:
Sends a message like this
"User morth1 has requested ownership of some replays. You can review this request at http://localhost:8000/users/claim/17 "
to discord when a new claim is created.

Using webhooks. If this is the better way to do things then I will close #402.


project/thscoreboard/users/forms.py
```diff
@@ -193,6 +193,7 @@ class ClaimReplaysForm(forms.Form):
         label=_("contact_info"), required=True, max_length=200
     )
     silentselene_username = forms.CharField(widget=forms.HiddenInput(), required=False)
+    submit = forms.CharField(widget=forms.HiddenInput(), required=False)
```
n-rook, line 196: This feels weird to me, what does this submit field actually do?

project/thscoreboard/users/test_claim.py
```diff
@@ -0,0 +1,255 @@
...
+        self.assertContains(response, "Success!")
+        claim_replay_request = user_models.ClaimReplayRequest.objects.first()
+        self.assertEqual(claim_replay_request.user, self.user)
+        self.assertEqual(claim_replay_request.contact_info, "contact info")
+        self.assertEqual(
+            claim_replay_request.request_status, user_models.RequestStatus.SUBMITTED
+        )
+
+    def test_staff_views_request_page(self) -> None:
+        if False:  # Disabled because it fails on github CI
```
n-rook, line 78: 1. Use @skip from https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures instead
2. If a test doesn't pass, I would just not include it in the first place

project/thscoreboard/users/views/claim.py
```diff
@@ -131,3 +201,16 @@ def _get_unclaimed_replays_from_username(
         .filter(user__isnull=True)
         .all()
     )
+
+
+def _assert_user_can_see_claim(
```
n-rook, line 206: In programming, an `assert` checks an invariant, something that should always be true. In theory, it should be possible to strip all asserts from a codebase with no changes, but in practice, they are there to catch bugs and warn about states that should be impossible. This is a long way for me to say "this function should probably be _check_user_can_see_claim".

project/thscoreboard/users/test_claim.py
```diff
@@ -0,0 +1,255 @@
+from django.db import connection
```
n-rook, line 1: This test should be in project/thscoreboard/users/views, right?

(This is cool, I've been wondering about how to test views well for awhile now. I think there's also a way to get views to return objects that are easier to assert about than strings, which doesn't test templates but might be easier for some other cases.)

project/thscoreboard/users/views/claim.py
```diff
@@ -76,11 +125,17 @@ def _create_new_claim_replay_request(
     )
     claim_replay_request.save()
     claim_replay_request.replays.set(replays)
+    discord_announce.announce(
```
n-rook, line 128: A concern: I think whatever channel this is, it probably should be private; I don't think people would expect their claims to be expressed to everyone. Obviously in most cases it won't matter (since when the claim is approved, obviously the replays will be attached to that person anyway) but I still think that'd be for the best.

n-rook (changes requested):
This change feels big, and I'm not sure how most of it relates directly to sending a discord message when a user creates a new claim replay request. I think you might be better off sending most of this as one PR, and then sending only the "discord webhook" part on a second PR.

jgdhs27:
> This change feels big, and I'm not sure how most of it relates directly to sending a discord message when a user creates a new claim replay request. I think you might be better off sending most of this as one PR, and then sending only the "discord webhook" part on a second PR.

Sorry, it should only be the final commit. The rest of the changes are from #400, which this depends on. Is there a good way to show that?

project/thscoreboard/users/forms.py
jgdhs27: When I parse the data by putting it into a form again, this field is used to track which button was pressed to submit this form. Could be "Approve", "Delete request", or "Confirm"

project/thscoreboard/users/test_claim.py
jgdhs27: i think its a nice test to have, i just cant figure out how to make it work on github ci. i'll try to skip it there only using skipIf

project/thscoreboard/users/views/claim.py
jgdhs27: done

project/thscoreboard/users/test_claim.py
jgdhs27: the test doesnt seem to be recognized by `manage.py test` if i put it there. (same issue is why i put test_import_royalflare to level)

project/thscoreboard/users/views/claim.py
jgdhs27: the requests themselves are private - only staff and the submitter can view them. the only thing we are giving away here is the username, i consider this to not be a privacy concern and its nice for the person reviewing it.

n-rook:
There's a trick I use to show a dependent PR: I set the merge target as the other PR it depends on. However I am not sure if there's a good way to do that when using a fork repo, since a PR from one of your repo's branches to another of your repo's branches would probably not actually show up here.

PR #404, opened 2023-05-18
Split large checkbox lists into columns of 4

jgdhs27:
Styling change. th08 and th09 have long shot lists, which occupy a large part of the screen. Therefore, I split them into columns of 4.
image
image

### DIFF
--- a/project/thscoreboard/replays/templates/replays/game_scoreboard.html
+++ b/project/thscoreboard/replays/templates/replays/game_scoreboard.html
@@ -4,7 +4,7 @@
     <h2><a href="/replays/{{game.game_id}}">{{ game.GetFullName }}</a></h2>
 
     {% for filter_name, filter_values in filters.items %}
-        <div class="checkbox-list">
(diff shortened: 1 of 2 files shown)


project/thscoreboard/replays/templates/replays/game_scoreboard.html
```diff
@@ -4,7 +4,7 @@
     <h2><a href="/replays/{{game.game_id}}">{{ game.GetFullName }}</a></h2>
 
     {% for filter_name, filter_values in filters.items %}
-        <div class="checkbox-list">
+        <div class="{% if filter_values|length >= 8 %}checkbox-list-4rows{% else %}checkbox-list{% endif %}">
```
n-rook, line 7: if you're feeling ambitious, I bet it's possible to do this purely in CSS, which would probably be better style. however this is also fine

n-rook (approved):

project/thscoreboard/replays/templates/replays/game_scoreboard.html
jgdhs27: I've already spent too much time trying to do it in a nicer way! 😅

PR #405, opened 2023-05-20
th128 support

32th-System:


jgdhs27:
general comment: please put things in numerical order whenever possible. it makes it a lot easier to find code

jgdhs27:
i tried it out and everything seems to work 👍

n-rook (changes requested):
Thanks for doing this! Scene games are kinda different and will require more work, but 12.8 really was a big hole in the stuff we supported, and it'll be good to have it. Only blocker is that (as morth mentioned), we should really put 12.8 between 12 and 13 in every file where we wind up listing every game in order.

project/thscoreboard/replays/game_ids.py
```diff
@@ -122,6 +123,11 @@ class NameLength(enum.Enum):
             pgettext("standard game name", "Undefined Fantastic Object"),
             pgettext("full game name", "東方星蓮船 - Undefined Fantastic Object"),
         ),
+        GameIDs.TH128: (
+            pgettext("short game name", "th128"),
```
jgdhs27, line 127: It should rather say "th12.8" on the front-end, not "th128"

jgdhs27 (approved):
Cool, just fix the short name used in the frontend

project/thscoreboard/replays/game_ids.py
32th-System: people ususally don't use the dot for game IDs for point releases

jgdhs27: I'm not sure where youre getting that info from, and Id be very surprised. royalflare uses 12.8, searching th128 on trs discord gets you to a sarcastic LeSanae poster...

32th-System: For example, on something like thpatch.net, or in thcrap they don't put the dot. The URL on Royalflare also doesn't have the dot

jgdhs27: dont mean the url, i mean the game name that's displayed on the front page of ss when you upload a replay

jgdhs27: thpatch.net uses the dot in this context

jgdhs27: image
just to be really clear, i am talking about here

n-rook: This is only showing up because you don't have TL set up locally--- on the production site, strings like "th17" shouldn't ever appear, and it says (for example) WBaWC instead. (Or something else for the JPN site.) So I think it's OK to say th128.

n-rook (approved):

PR #407, opened 2023-05-22
Update django.po

32th-System:


project/thscoreboard/users/forms.py
```diff
@@ -189,9 +189,7 @@ class ClaimReplaysForm(forms.Form):
     choices = forms.MultipleChoiceField(
         widget=forms.CheckboxSelectMultiple,
     )
-    contact_info = forms.CharField(
-        label=_("contact_info"), required=True, max_length=200
-    )
+    contact_info = forms.CharField(required=True, max_length=200)
```
n-rook, line 192: This change is unrelated

project/thscoreboard/locale/ja/LC_MESSAGES/django.po
```diff
@@ -1,1897 +1,2415 @@
...
-msgstr ""
-
-#: users/templates/users/profile.html:5
-msgid "My Profile"
-msgstr "プロフィール"
-
-#: users/templates/users/profile.html:41
-msgid "Log out"
-msgstr "ログアウト"
+# Silent Selene translation file (in gettext .po format).
```
n-rook, line 1: This is on me, but I actually think we shouldn't update the ja django.po file ourselves anymore. We should just let Transifex handle it. My understanding is that Transifex only reads the English .po file from our repo, so I think changes to this file will just get overwritten.

n-rook (approved):
looks good pending comments

PR #410, opened 2023-05-25
Check for th128 before checking for stage == 7

32th-System:
Fixes https://www.silentselene.net/replays/th128/24391


jgdhs27 (changes requested):
Please update the corresponding test

jgdhs27:
> this part of the code isn't covered by tests

See `FormatStageTestCase` project\thscoreboard\replays\test_game_fields.py

Please add at least a test that would fail before this change, and passes after this change.

PR #411, opened 2023-05-25
th128: add frozen area

32th-System:


project/thscoreboard/replays/game_fields.py
```diff
@@ -30,6 +30,8 @@
         "th09_p2_score": False,
         "th128_motivation": False,
         "th128_perfect_freeze": False,
+        "th128_frozen_area": False,
+        "th128_frozen_area": False,
```
jgdhs27, line 34: repeated

project/thscoreboard/replays/game_fields.py
```diff
@@ -642,6 +658,11 @@ def FormatStages(game_id: str, replay_stages: Iterable[models.ReplayStage], shot
         else:
             stage.th128_perfect_freeze = f"{stage.th128_perfect_freeze//100}%"
 
+        if stage.th128_frozen_area is None:
+            stage.th128_frozen_area = ""
+        else:
+            stage.th128_frozen_area = f"{int(stage.th128_frozen_area)}%"
```
jgdhs27, line 664: Why are the casts here and in test_replay_parsing.py necessary? According to the `ReplayStage` dataclass, this is already an int

jgdhs27: This is still an IntegerField according to models.py so something is still off.

project/thscoreboard/replays/game_fields.py
```diff
@@ -642,6 +657,11 @@ def FormatStages(game_id: str, replay_stages: Iterable[models.ReplayStage], shot
...
```
jgdhs27, line 663: also, if this is a conversion from float to percentage it looks very strange.uUsually such conversions have a "/100" somewhere. if zun really did use a float value of 100 to represent 100%, consider adding a comment to explain it, because it's really surprising.

jgdhs27 (changes requested):

project/thscoreboard/replays/game_fields.py
jgdhs27: actually, isnt it usually 100? and the above values use a different convention where 10,000.0 represents 100%? i might be missing something but this looks very strange

32th-System: Yeah the above percentage is a fixed point integer, while this is a floating point number where 100% is 100. (And where anything after the decimal point gets cut off)

jgdhs27 (approved):
looks good to me

PR #412, opened 2023-05-29
Only show 'Claim replays from royalflare' link on own profile

jgdhs27:
Bugfix. The 'Claim replays from royalflare' link at the bottom of the user profile should only be seen when viewing your own profile.


32th-System (approved):
I approve but maybe that link should (also) go on /users/profile?

jgdhs27:
I have the impression that users/profile is more of an "account settings" page and this wouldn't really fit. Now that I think of it, maybe that page should be renamed.

PR #413, opened 2023-05-29
Update requirenments.txt

32th-System:
Dependabot is complaining about a security issue and Morth wants to use Django 4.2 for #360


n-rook (approved):
Be sure to click around on the site to make sure it works before submitting

PR #415, opened 2023-06-01
Create player rankings page

jgdhs27:
Create a new player rankings page that shows player's ranks (1st place, 2nd place & third place count) across selected games. A user may pick a specific game, all windows/pc-98 games, or all games.

image

### DIFF
--- a/project/thscoreboard/replays/forms.py
+++ b/project/thscoreboard/replays/forms.py
@@ -7,6 +7,7 @@ from django import forms
 from django.core import exceptions
 from django.db.models import enums
 
+from replays.get_all_games import get_all_games_by_category
 from replays import game_ids
 from replays import game_fields
 from replays import models
@@ -193,6 +194,7 @@ class EditReplayForm(forms.Form):
 
 class RankingGameSelectionForm(forms.Form):
     all_games = models.Game.objects.all()
-    game_selection = forms.ChoiceField(
-        choices=[(game.game_id, game.GetShortName()) for game in all_games]
-    )
+    all_game_names = [game.GetShortName() for game in all_games]
+    all_choices = all_game_names + ["All games", "Windows", "PC-98"]
+    print(all_choices)
+    game_selection = forms.ChoiceField(choices=[(c, c) for c in all_choices])
--- a/project/thscoreboard/replays/forms.py
+++ b/project/thscoreboard/replays/forms.py
@@ -7,7 +7,7 @@ from django import forms
 from django.core import exceptions
 from django.db.models import enums
 
-from replays.get_all_games import get_all_games_by_category
+from replays.get_all_games import get_pc98_games, get_windows_games
 from replays import game_ids
 from replays import game_fields
 from replays import models
@@ -193,8 +193,30 @@ class EditReplayForm(forms.Form):
 
 
 class RankingGameSelectionForm(forms.Form):
-    all_games = models.Game.objects.all()
-    all_game_names = [game.GetShortName() for game in all_games]
-    all_choices = all_game_names + ["All games", "Windows", "PC-98"]
-    print(all_choices)
-    game_selection = forms.ChoiceField(choices=[(c, c) for c in all_choices])
+    SELECT_ALL = "All games"
+    SELECT_PC98 = "PC-98"
+    SELECT_WINDOWS = "Windows"
+    GROUPED_SELECTIONS = [SELECT_ALL, SELECT_PC98, SELECT_WINDOWS]
+    DEFAULT_SELECTION = SELECT_WINDOWS
+
+    grouped_game_selection = forms.ChoiceField(
+        choices=[(g, g) for g in GROUPED_SELECTIONS], required=False
+    )
+    pc98_game_selection = forms.ChoiceField(
+        choices=[(c.GetShortName(), c.GetShortName()) for c in get_pc98_games()],
+        required=False,
+    )
+    windows_game_selection = forms.ChoiceField(
+        choices=[(c.GetShortName(), c.GetShortName()) for c in get_windows_games()],
+        required=False,
+    )
+
+    def get_selection(self) -> str:
+        print(self.cleaned_data)
+        print(self.cleaned_data["windows_game_selection"])
+        if "grouped_game_selection" in self.cleaned_data:
+            return self.cleaned_data["grouped_game_selection"]
+        if "pc98_game_selection" in self.cleaned_data:
+            return self.cleaned_data["pc98_game_selection"]
+        if "windows_game_selection" in self.cleaned_data:
+            return self.cleaned_data["windows_game_selection"]
--- a/project/thscoreboard/replays/forms.py
+++ b/project/thscoreboard/replays/forms.py
@@ -211,6 +211,14 @@ class RankingGameSelectionForm(forms.Form):
         required=False,
     )
 
+    def get_selection(self) -> str:
+        return (
+            self.cleaned_data["grouped_game_selection"]
+            or self.cleaned_data["pc98_game_selection"]
+            or self.cleaned_data["windows_game_selection"]
+            or self.DEFAULT_SELECTION
+        )
+
     all_choice_fields = [
         ("grouped_game_selection", grouped_game_selection),
         ("pc98_game_selection", pc98_game_selection),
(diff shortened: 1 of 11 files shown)


project/thscoreboard/replays/get_all_games.py
```diff
@@ -0,0 +1,29 @@
+from typing import Iterable
+from replays import game_ids
+
+from replays import models
+
+
+PC98_GAME_IDS = [
```
n-rook, line 7: I won't insist on it in this PR, because it'd require some restructuring of existing code and be quite the project, but it would really be best if this metadata were actually in the database, rather than having us hardcode it like this.

project/thscoreboard/replays/views/replay_list.py
```diff
@@ -111,7 +112,7 @@ def _get_filter_options_th17(game: Game) -> dict[str, list[str]]:
     }
 
 
-def _get_all_replay_for_game(game_id: str) -> dict:
+def get_all_replay_for_game(game_id: str) -> Manager[models.Replay]:
```
n-rook, line 115: As a general rule, views should not import code from other views. Rather, if we have queries like this which are useful in multiple places, we should put them in a single location (maybe in models.py, maybe in a separate location). After all, nobody would go looking in the views directory to find useful functions that might save them time.

project/thscoreboard/users/views/rankings.py
```diff
@@ -0,0 +1,137 @@
...
+            self.first_place_count,
+            self.second_place_count,
+            self.third_place_count,
+        ) < (other.first_place_count, other.second_place_count, other.third_place_count)
+
+
+@http_decorators.require_http_methods(["GET", "HEAD", "POST"])
+def rankings(request: WSGIRequest) -> HttpResponse:
+    selection = forms.RankingGameSelectionForm.DEFAULT_SELECTION
+    if request.method == "POST":
```
n-rook, line 34: I don't think we should use POST here. It's nice to use GET rather than POST for queries that don't change the database, like this one: most browsers will prompt before refreshing a POST, which is inconvenient for users. Plus, since the POST data isn't present in the URL, a link to it can't be shared. I've never actually written a GET form with Django, but it looks like it supports them cleanly (https://docs.djangoproject.com/en/4.2/topics/forms/#get-and-post) so I think this will not be that hard to change.

project/thscoreboard/users/views/rankings.py
```diff
@@ -0,0 +1,137 @@
...
+
+def _get_all_ranked_categories_for_game(
+    game: replay_models.Game,
+) -> Iterable[tuple[replay_models.Shot, int, replay_models.Route]]:
+    # Excludes unranked categories, such as replays that have no route due to early game-overs.
+
+    all_shots = replay_models.Shot.objects.filter(game=game)
+    all_difficulties = list(range(game.num_difficulties))
+    all_routes = replay_models.Route.objects.filter(game=game)
+    if len(all_routes) == 0:
```
n-rook, line 98: Can't you just do `if not all_routes` here, and similar for the other checks? That feels more idiomatic in Python.

project/thscoreboard/replays/forms.py
```diff
@@ -189,3 +190,29 @@ def __init__(self, *args, game: models.Game, **kwargs):
 
 class EditReplayForm(forms.Form):
     comment = _create_comment_field()
+
+
+class RankingGameSelectionForm(forms.Form):
```
n-rook, line 195: I think this class is a good place to encapsulate the question of "which games should we display". Define a method (something like GetSelectedGameIDs) that will return the selected games, and handle that logic there. Currently, all of this is handled in the view, and you wind up with some circuitous logic there.

project/thscoreboard/replays/forms.py
```diff
@@ -189,3 +190,29 @@ def __init__(self, *args, game: models.Game, **kwargs):
...
+    SELECT_ALL = "All games"
+    SELECT_PC98 = "PC-98"
+    SELECT_WINDOWS = "Windows"
+    GROUPED_SELECTIONS = [SELECT_ALL, SELECT_PC98, SELECT_WINDOWS]
+    DEFAULT_SELECTION = SELECT_WINDOWS
```
n-rook, line 200: Someday I will complain about this... but for now there are not enough uploads for pc98 games so the data there probably looks dumb. So I'll allow it for now :p

project/thscoreboard/users/views/rankings.py
```diff
@@ -0,0 +1,137 @@
...
+        },
+    )
+
+
+def _get_all_player_rankings_for_games(
+    games: Iterable[replay_models.Game],
+) -> dict[str | user_models.User, RankCount]:
+    rankings = defaultdict(lambda: RankCount(0, 0, 0))
+    for game in games:
+        all_replays = get_all_replay_for_game(game)
```
n-rook, line 63: This is going to be a pain in the ass request, and I'm sorry for making it, but I think it's worth doing: Please verify, by inspecting the SQL query used by Django (with something like https://stackoverflow.com/questions/4375784/how-to-log-all-sql-queries-in-django), that we do not fetch every replay in this function. If we fetch every replay, this function will be kinda slow, but we really shouldn't need to for this problem.

n-rook (changes requested):
This is very cool

project/thscoreboard/users/views/rankings.py
n-rook: Actually, wait, rather than this, you should really just figure out how to get this data directly from the database without having to do all of this computation in Python. Like, this is the sort of problem that databases are good at solving. If you figure out how to query this properly, it should be very fast. However, you will probably have to either use complicated ORM options or just give up and use raw SQL (https://docs.djangoproject.com/en/4.2/topics/db/sql/#performing-raw-queries).

project/thscoreboard/replays/get_all_games.py
jgdhs27: That makes sense. I might tackle that later, for now i'll make an issue

project/thscoreboard/replays/views/replay_list.py
jgdhs27: makes sense. in fact this query is probably doing bad performance-wise anyway, so use a new one instead

project/thscoreboard/users/views/rankings.py
jgdhs27: I'm looking into it and I think Django 4.2's filter-after-window is again perfect for this

jgdhs27: Alright, it works like a char,

project/thscoreboard/replays/forms.py
jgdhs27: done

jgdhs27:
It's ready for another review. I have no idea what's causing the CI failure:
`django.db.utils.OperationalError: connection to server at "localhost" (::1), port 5432 failed: FATAL:  database "thscoreboard" does not exist`

32th-System:
maybe me merging the requirenments.txt update? Try merging main into this branch

jgdhs27:
Figured out the issue - I was doing some database queries top-level (TIL class constants are top-level).

But I was playing around one last time and think the calculations for all games are incorrect, so I still need to fix that

n-rook:
could you close the PR if it's not ready to review right now? not that this is a big deal, but that way I don't have to worry that I'm letting an open PR languish when I should be reviewing it

jgdhs27:
ok

jgdhs27:
Re-opening this PR. I looked into the ranking weirdness and think I solved it in #435.

project/thscoreboard/shared_content/static/themes.sass
```diff
@@ -40,7 +40,9 @@
     // for the rows.
     --table-header-color: #ffb3ff
     --table-row-odd-color: #fff3ff
-    --table-row-even-color: #ffe6ff
```
n-rook, line 43: It looks like this was deleted by accident, no?

n-rook (approved):
Looks good! Just one comment

project/thscoreboard/shared_content/static/themes.sass
jgdhs27: Oops! Re-adding it

PR #426, opened 2023-06-17
fix tickbox behaviour if page gets loaded with a box ticked

32th-System:
This happens when the page gets loaded from going back from a replay page, or when reloading


jgdhs27:
This doesn't fix the issue for me.
I am testing on edge btw, it seems to have the bug most consistently.

32th-System:
I tested in Firefox, I'll look into it. Maybe the load event isn't being triggered?

32th-System:
In Chromium, reloading doesn't leave the box ticked, and going back uses a cached version of the page that has the table fully intact

32th-System:
turns out Edge handles things a bit differently from Chromium sometimes. The correct event to use was `pageshow`

jgdhs27 (approved):
Cool 👍

jgdhs27:
@32th-System could you please rebase/merge and re-run the tests? I've removed them for being flaky

PR #431, opened 2023-07-01
Revamp the replay details page.

n-rook:
There's still work to do on this page, but I think it's already an improvement.

Screen Shot 2023-07-01 at 1 57 33 PM

The main goal of this redesign was to introduce more visual hierarchy on the page. In other words, I want more important or interesting information to be clearer and closer to the top of the page. That's why I added the main score box, which makes the score jump out and also presents the most interesting metadata, as well as a simpler replay details table which includes only the info most interesting for scoring runs (score) and survival runs (lives and bombs).

Known issues with the page:
- It doesn't look great for replays other than full game runs of Windows games.
- The SVG icons used as badges are ugly.
- The css on the cards themselves could be improved; I don't love the way they look right now. I tried a different background color but couldn't find something I liked.
- Adding badges will likely make players more interested in editing their replay to fix the metadata if they didn't tag it as LNN (or whatever), but we don't have that feature yet. The same is true if and when we add "no third category"-type annotations.


project/thscoreboard/replays/templates/replays/replay_details.html
```diff
@@ -11,121 +12,206 @@
...
+        {% endif %}
+    </div>
+
+    <div class="replay-box replay-spiffy-table-parent">
+        <div class="replay-badges">
+            {% if replay.is_clear %}
+            <div class="replay-badge">
+                <svg fill="currentColor">
+                    <title>This run cleared the game</title>
+                    <use href="{% static 'icons/bootstrap/icons.svg' %}#circle" />
```
jgdhs27, line 92: 99% of replays will just have a circle, and its meaning is unclear. I'm not sure about this. I think the NB and NM icons are a nice idea though

project/thscoreboard/replays/templates/replays/replay_details.html
```diff
@@ -11,121 +12,206 @@
...
+                <svg fill="currentColor">
+                    <title>This run never missed</title>
+                    <use href="{% static 'icons/bootstrap/icons.svg' %}#slash" />
+                    <use href="{% static 'icons/bootstrap/icons.svg' %}#heart" />
+                </svg>
+            </div>
+            {% endif %}
+        </div>
+
+        {% if replay_stages %}
```
jgdhs27, line 116: this looks a bit strange on Ex replays with only one stage. I think it would be better with stage names and/or a header like "stage splits".

project/thscoreboard/replays/templates/replays/replay_details.html
```diff
@@ -11,121 +12,206 @@
...
-                <td>Spell Card</td>
-                <td>{{ spell_name }}</td>
+                {% if table_fields.lives %}
+                <td class="replay-lives-column">
+                    {% if stage.lives %}
+                    {{ stage.lives }}
+                    <svg class="inline-icon">
+                        <use href="{% static 'icons/bootstrap/icons.svg' %}#heart" />
+                    </svg>
+                    {% endif %}
```
jgdhs27, line 127: this is really cool

project/thscoreboard/replays/templates/replays/replay_details.html
```diff
@@ -11,121 +12,206 @@
...
+
+    {% if has_stages %}
+    <div class="replay-box">
+        <h3>Stage details</h3>
+        {% include "replays/replay_stages.html" %}
     </div>
     {% endif %}
-    {% if can_edit %}
+    <div class="replay-box">
+        <h3>Edit replay</h3>
```
jgdhs27, line 208: Not sure about "Edit" here. Maybe "Manage" would fit better?

jgdhs27 (changes requested):
Sorry, but could you please improve the styling? It would be nice to have some more colors. Also, the contrast between the thick white lines and the background on dark mode is too high.

32th-System:
I'm not sure how much I like that the box at the top right shows the same information as the "Stage details" box. Maybe all the stage properties should be converted to use icons, or Stage details should have some of the information taken out. I'm also not sure about showing lifes specifically in the top right box, since some games don't have a "lifes" system to speak of.

What I do however like, is how this could look with showing the cards in th18 or the items in th143

project/thscoreboard/replays/templates/replays/replay_details.html
```diff
@@ -11,121 +12,206 @@
...
+            {{ replay.score|intcomma }}
+        </p>
+        <p>
+            {% if replay.timestamp %}
+              {% with replay_timestamp=replay.GetFormattedTimestampDate %}
+                {% if owned %}
+                  {% with username=replay.user.username %}
+                    {% url 'user_page' username as user_url %}
+                    {% blocktranslate %}
+                        Played by <a href="{{ user_url }}">{{ username }}</a> on {{ replay_timestamp }}
```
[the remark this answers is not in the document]
n-rook: I agree, but sadly there is another reason I preferred it this way: the translation. The whole block is given to translators. If I move `on {{ replay_timestamp }}` out of the block, I'll have to put it in a different {% blocktranslate %} tag, which means the translators will have to tl "Played by …" and "on …" independently, which seems like it would be harder.

project/thscoreboard/replays/templates/replays/replay_details.html
n-rook: I didn't like it much either, but couldn't think of anything better. I think Manage is at least slightly better; I'll try that

n-rook:
The top box does show bombs too if that's information we have for the game; it just isn't in the screenshot above because it's mof. eg

Screenshot 2023-07-23 at 4 14 36 PM

It's not in this PR, but I figured we could show motivation and perfect freeze for GFW. I considered putting PIV in there too but decided not to, just because it's irrelevant for survival games (whereas on the other hand, resource counts are still interesting for most scoring runs).

Having two tables is not ideal, but I am not sure a better way to handle it. raviddog added a bunch of information collected from replays that most people won't care about, so I wanted to present the most interesting data front and center. But it would be a shame to hide that data entirely from the site; some of it is still interesting, and at the very least, it's cool that we have it. We could put it on another web page entirely, but that seemed like too big of a task for this PR, and I'm not sold on it as an idea in the first place.

A bigger deal (and something I'm not sure about) is what to do with runs other than full game runs of Windows games. That's a definite minority of runs, so I kinda figured we could merge this PR and then improve the situation later.

I totally agree that a good part about this layout is that we can add boxes for specific things (like UM cards) that wouldn't be relevant for most runs.

project/thscoreboard/replays/templates/replays/replay_details.html
n-rook: Something else I actually wanted to do was to add icons for unfinished runs too— ⑥ etc. But I figured I'd do it later.

The issue we have is that we have to put this information _somewhere_, and it should probably be somewhere prominent. After all, NM and NB are the most interesting things about replays where they're the case, and similarly, it's usually really relevant to know if a replay cleared or not. The good thing about the icons is that they only take up space when they're active (we don't need to have that goofy "Used bombs? Yes" on almost every run). The circle is meant to evoke success, but I agree that it's not great; I had hoped to solicit SVG icons later (or try my own extreme art skills, I guess).

n-rook:
I've been putting off responding to this because css scary, but I'll try to put some time away to try other colors out.

n-rook:
Screenshot 2023-07-29 at 1 58 06 PM
Added some light background and shadow elements, what do you think? (note: not uploaded to this PR yet)

n-rook:
oh I should upload a screenshot of the whole page, not just the replay details part, because I've found screenshots of just one part look unnaturally dreary because they miss the banner and borders

Screenshot 2023-07-29 at 2 00 04 PM

n-rook:
here's an attempt at a dark mode style— I think it looks better than the one with the white borders, at least

Screenshot 2023-07-29 at 2 09 55 PM

jgdhs27:
The purple highlights on dark mode help so much! It fits a lot better with the site's theme. I still think that light mode feels too monochromatic.

project/thscoreboard/replays/templates/replays/replay_details.html
n-rook: Thanks!

n-rook (commented):
I updated the light mode to have colored borders as well. I think this is ready for review again; please let me know what you think!

n-rook:
Screenshot 2023-07-30 at 5 09 30 PM

jgdhs27 (approved):
I love this! It looks neat and all the right info sticks out of the page 😄

n-rook:
Per popular demand, I got rid of the clear icon :)

I'll try to bring it back at some point, of course :p But hopefully with a different icon.

PR #435, opened 2023-07-28
Sort tied replays by upload time

jgdhs27:
In case of ties, which happen frequently in counterstopped games, rank replays based on upload date. If that doesnt fix the tie (because 4 players counterstop on day 1, thanks zun), sort by ID as a desperate last resort to make rankings deterministic.


n-rook (approved):
Seems reasonable

PR #436, opened 2023-07-29
Remove test_claim.py

jgdhs27:
These tests are interfering with each other and other tests.
Making every single test its own transaction seemed to fix that, but then the tests are too slow.


n-rook:
Hmm. Sad, but fair— I've also seen flaky test failures coming from these tests.

I'll approve this, but I feel like the root cause here must be a bug, either in the test code or in our real server somewhere. Tests run serially, so in theory even these tests should never be interfering with other ones.

n-rook (approved):

PR #440, opened 2023-08-01
Rework leaderboard

jgdhs27:
image

- Replaces checkboxes with nice big buttons. Much prettier and easier to press!
- Only allow 0-1 option in each category to be selected (eg. disallow selecting both Reimu + Marisa at once)
  - This removes a feature, but saves you a click when you want to switch to just one category to another
  - It's also preparation for having simple url query strings for each category, like replays/th08?shot=youmu&route=final%20b

### DIFF
--- a/project/thscoreboard/replays/templates/replays/game_scoreboard.html
+++ b/project/thscoreboard/replays/templates/replays/game_scoreboard.html
@@ -4,7 +4,7 @@
     <h2><a href="/replays/{{game.game_id}}">{{ game.GetFullName }}</a></h2>
 
     {% for filter_name, filter_values in filters.items %}
-        <div class="{% if filter_values|length >= 8 %}checkbox-list-4rows{% else %}checkbox-list{% endif %}">
+        <div class=button-container>
             {% for filter_value in filter_values %}
                 <label><button
                     class="button"
--- a/project/thscoreboard/replays/templates/replays/game_scoreboard.html
+++ b/project/thscoreboard/replays/templates/replays/game_scoreboard.html
@@ -5,6 +5,12 @@
 
     {% for filter_name, filter_values in filters.items %}
         <div class=button-container>
+            <label><button
+                class="button pressed"
+                filterType="{{filter_name}}"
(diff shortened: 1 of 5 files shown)


n-rook (changes requested):
Nice, I mostly like it.

I'm concerned about the integration of the UX concept that clicking an active filter disables it. We don't do that on the leaderboard, and I think it would be hard for users to discover. I'd much prefer having an "All" button, as we do on the leaderboard.

A minor nitpick: On Chrome, if I hover between two buttons, I see an I-beam. I'm not sure why, but it'd be nice to get rid of that if you can figure it out.

jgdhs27:
> I'm concerned about the integration of the UX concept that clicking an active filter disables it. We don't do that on the leaderboard, and I think it would be hard for users to discover. I'd much prefer having an "All" button, as we do on the leaderboard.

How do I set "All" up for translators? Is there a shared context? Should I invent a new one?

n-rook:
In theory, you should invent a new one, with a description like "selecting all of a category".

I actually think this PR would be much simpler if you handled the "All" button exclusively on the frontend / js side. That is, had a button like the current filters, except it calls "onClickAll()" or something instead. That way you wouldn't have to make backend changes or pass the literal string "All" around anywhere. It's kind of redundant to be manually prepending the string "All" to all these different filter lists fetched from the database anyway.

jgdhs27:
You're right, I was already working on having initial filters for the next PR, but that's overdoing it for now. This new version should be simpler.

jgdhs27:
> Nice, I mostly like it.
> 
> I'm concerned about the integration of the UX concept that clicking an active filter disables it. We don't do that on the leaderboard, and I think it would be hard for users to discover. I'd much prefer having an "All" button, as we do on the leaderboard.
> 
> A minor nitpick: On Chrome, if I hover between two buttons, I see an I-beam. I'm not sure why, but it'd be nice to get rid of that if you can figure it out.
Can't reproduce

n-rook (approved):

PR #442, opened 2023-08-05
Add replay-specific meta tags.

n-rook:
This way, links to replays will look better in places like Discord.

The templating unfortunately does have to be a horrible never-ending line, because whitespace inside meta tags modifies how it is rendered.
Screenshot 2023-08-05 at 3 39 02 PM

### DIFF
--- a/project/thscoreboard/replays/templates/replays/replay_details.html
+++ b/project/thscoreboard/replays/templates/replays/replay_details.html
@@ -11,6 +11,8 @@
 {% endif %}
 {% endblock %}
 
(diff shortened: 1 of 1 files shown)


jgdhs27:
How do I test this? I tried checking out the PR and posting a link like http://localhost:8000/replays/th14/299950 on discord, didnt get any results

jgdhs27 (approved):
Oh, I see the example htmls and the tests on discord, and confirmed i see the same html as those examples

n-rook:
yeah there are basically two ways to do it

way number one is to just copy paste the meta tags into somewhere like a github gist that discord will render. that's what I did
way number two (the "real" way) is to use a service like ngrok to tunnel your local server to the open internet. this is honestly better for actual testing, but you have to actually do it of course

PR #444, opened 2023-08-06
Embed videos

32th-System:


project/thscoreboard/replays/views/view_replay.py
```diff
@@ -23,6 +26,60 @@
 from replays import game_ids
 
 
+def _get_twitch_embed_link(split_url: ParseResult):
```
n-rook, line 29: We should put the embed functions in a new module instead. There are two reasons for this:

1. This module is already pretty long and confusing.
2. The API will make more sense this way. Right now, get_video_embed_link is public, but it's "really" private— we wouldn't want people importing view_replay in order to use it. The view modules in general should only really be imported by URL handlers and things like that. But a function that turns a video URL into an embed link is really nice and makes sense as a public function, so it makes sense to put it in a separate function so it can really be public.

project/thscoreboard/replays/test_view_replay.py
```diff
@@ -47,3 +50,69 @@ def testFields(self):
                                 self.assertIsNone(
                                     s[key], msg=f"Unexpected field {key} found"
                                 )
+
+
+class VideoEmbedTestCase(unittest.TestCase):
```
n-rook, line 55: Split these tests up into one assertion per method. That way, if we break (say) youtu.be embeds it won't short-circuit the test and prevent us from knowing if we broke something else.

project/thscoreboard/replays/views/view_replay.py
```diff
@@ -23,6 +26,60 @@
...
+
+
+def _get_bilibili_embed_link(split_url: ParseResult):
+    path = split_url.path.strip("/").split("/")
+    if path[0] == "video" and len(path) == 2:
+        video_id = path[1]
+        return f"https://player.bilibili.com/player.html?bvid={video_id}&autoplay=false"
+
+
+@lru_cache(maxsize=None)
```
n-rook, line 64: I think I found a memory leak :p

Other than the memory issue (which will almost certainly not be a problem, honestly, but we should still set some sort of max size), I do wonder about the necessity of this. These calculations are really cheap; we're just doing string editing. So I don't think we need a cache. (And remember that even now we are running lots of copies of the server with WSGI. So cache hits will be rare even if someone looks at the same page repeatedly)

project/thscoreboard/replays/templates/replays/replay_details.html
```diff
@@ -146,6 +146,14 @@ <h3>Badges and Stage Splits</h3>
         {% endif %}
     </div>
```
n-rook, line 148: If we have a video embed, we can get rid of the video link in the Score pane, right?

project/thscoreboard/shared_content/static/style.sass
```diff
@@ -302,6 +302,17 @@ th, td
         margin-inline: 0
         margin-bottom: .7em
 
+.video-wrapper
+    padding-bottom: 56.25%
```
n-rook, line 306: This is very clever. IMO add a comment (.5625 = 9 / 16, maybe), because it's nonobvious even for CSS where that number comes from.

project/thscoreboard/replays/templates/replays/replay_details.html
32th-System: Not sure about that. The score pane has the link to the user in it. Also, depending on how long the embed takes to load, how the embed displays the video title, or if the video title has the score in it at all, having the score pane still there is still useful to me

jgdhs27:
whats with the Thank-you-morth? lol

project/thscoreboard/replays/lib/video_embed.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def _get_bilibili_embed_link(split_url: ParseResult):
+    path = split_url.path.strip("/").split("/")
+    if path[0] == "video" and len(path) == 2:
+        video_id = path[1]
+        return f"https://player.bilibili.com/player.html?bvid={video_id}&autoplay=false"
+
+
+def get_video_embed_link(url: str):
```
n-rook, line 39: Add a docstring! It can be short, but be sure to mention that None is returned if we can't construct a valid embed.

project/thscoreboard/replays/templates/replays/replay_details.html
n-rook: Oh, I just meant literally delete the link to the video alone, not the whole score pane.

n-rook (approved):
Nice!

project/thscoreboard/replays/templates/replays/replay_details.html
32th-System: I was just thinking: what if the embed doesn't load or takes really long to load? This happens to me with Bilibili videos for example, so having the link to the video somewhere else would be nice. I also have an issue where in my main browser, Bilibili won't play at all no matter if it loads or not, so I'd like to be able to copy the link and put it in another browser

PR #447, opened 2023-08-13
introduce typescript

jgdhs27:
Adding types here to make the logic easier to follow


project/thscoreboard/replays/templates/replays/replay_table.html
```diff
@@ -42,6 +42,7 @@
 <script>
     const showGameColumn = "{{show_game}}";
     const showRouteColumn = "{{show_route}}";
+    const aaaa = "aaaaa";
```
n-rook, line 45: ?

project/thscoreboard/shared_content/static/js/replays/replayTable.js
```diff
@@ -1,205 +1,200 @@
+"use strict";
```
n-rook, line 1: Sorry to be annoying by saying this, but: We really shouldn't check in generated code. Rather, we should recreate it whenever we deploy.

If you really need to, it might be okay to check it in as an intermediate step while you're trying to get the ts deploy working, but if so, please put a giant comment at the top warning that this is generated code. But it'd be way better to do it in one go.

project/thscoreboard/shared_content/static/js/replays/replayTable.ts
```diff
@@ -0,0 +1,247 @@
...
+
+try {
+  // @ts-ignore
+  module.exports = {updateFilters, filterReplays};
+} catch(error) {
+  // Shut up error in console but do nothing
+}
+
+function ensureTickboxSync(): void {
+  activeFilters = initializeFilters();
```
n-rook, line 239: Er--- why are we calling initializeFilters() three times in this module?

project/thscoreboard/replays/replays_to_json.py
```diff
@@ -23,38 +23,38 @@ def convert_replay_to_dict(self, replay: models.Replay) -> dict:
         score_prefix = _get_medal_emoji(replay.rank) if hasattr(replay, "rank") else ""
 
         json_dict = {}
-        json_dict["Id"] = replay.id
+        json_dict["id"] = replay.id
```
n-rook, line 26: IMO you should really make these changes in a separate PR; they make this one look much bigger than it needs to be

n-rook (changes requested):
Nice; using ts will be really convenient and make it easier to maintain the frontend without getting lost in the complexity of it.

PR #452, opened 2023-09-06
Add translate to most html files

jgdhs27:
Add translate blocks wherever its easy to do so.


n-rook (approved):
Thanks! This is definitely good to do

PR #468, opened 2023-09-24
Seamlessly delete replays that duplicate a new replay but are owned by an inactive user.

### LINKED ISSUE #462: Unable to upload replay that was previously uploaded from an account marked for deletion
To reproduce:
1. Upload a replay from an account
2. Delete the account (Mark it for deletion)
3. Upload the replay from another account

You will get am error message like
"This replay already exists. /replays/th11/321943"
However, trying to access the replay gives a 404

n-rook:
Fixes #462

### DIFF
--- a/project/thscoreboard/replays/create_replay.py
+++ b/project/thscoreboard/replays/create_replay.py
@@ -5,13 +5,53 @@ not with web entities like views or forms.
 """
 
 import datetime
+import logging
 from typing import Optional
-from django.db import transaction
 
+from django.db import transaction
+from django.db import utils
+
+from shared_content import db_errors
 from replays import constant_helpers
 from replays import game_ids
 from replays import models
 from replays import replay_parsing
+
+
+def _SaveNewReplayWithFile(r: models.Replay, rf: models.ReplayFile):
+    r.save()
+    try:
+        # If there is an error in a transaction, that transaction is rolled
+        # back. As such, we execute rf.save() in a nested transaction— that
+        # way, only this tiny transaction is rolled back, and the outer transaction
+        # (if any) can continue as expected. (Of course, if the error can't be
+        # handled here, it will be reraised, so the outer transaction will be
+        # rolled back too.)
+        with transaction.atomic():
+            rf.save()
+    except utils.IntegrityError as e:
+        if (
+            not db_errors.IsUniqueError(e)
+            or not db_errors.GetUniqueConstraintCause(e) == "unique_hash"
+        ):
+            raise
+
+        at_least_one_ghost = False
+        for ghost in models.Replay.objects.ghosts_of(rf.replay_hash):
+            logging.info(
+                "Deleting ghost replay (with ID %d; owned by %s)",
+                ghost.id,
+                ghost.user.username,
+            )
+            ghost.delete()
+            at_least_one_ghost = True
+
+        if at_least_one_ghost:
+            rf.save()
+        else:
+            # As there were no ghosts, the save attempt was blocked by
+            # a real duplicate.
+            raise
 
 
 @transaction.atomic
@@ -91,8 +131,7 @@ def PublishNewReplay(
         ),
     )
 
-    replay_instance.save()
-    replay_file_instance.save()
+    _SaveNewReplayWithFile(replay_instance, replay_file_instance)
     temp_replay_instance.delete()
 
     for s in replay_info.stages:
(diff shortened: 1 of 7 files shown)


project/thscoreboard/replays/create_replay.py
```diff
@@ -5,15 +5,55 @@
...
 
+from shared_content import db_errors
 from replays import constant_helpers
 from replays import game_ids
 from replays import models
 from replays import replay_parsing
 
 
+def _SaveNewReplayWithFile(r: models.Replay, rf: models.ReplayFile):
+    r.save()
```
jgdhs27, line 22: im not sure this should be outside the transaction. if there's a non-ghost duplicate, doesnt this mean that the replay saves without the replayfile?

maybe there's an assumption that this function always called within a transaction, but that feels dangerous.

project/thscoreboard/replays/create_replay.py
```diff
@@ -5,15 +5,55 @@
...
+        # way, only this tiny transaction is rolled back, and the outer transaction
+        # (if any) can continue as expected. (Of course, if the error can't be
+        # handled here, it will be reraised, so the outer transaction will be
+        # rolled back too.)
+        with transaction.atomic():
+            rf.save()
+    except utils.IntegrityError as e:
+        if (
+            not db_errors.IsUniqueError(e)
+            or not db_errors.GetUniqueConstraintCause(e) == "unique_hash"
```
jgdhs27, line 35: i feel like `"unique_hash"` should be extracted to a constant

project/thscoreboard/shared_content/test_db_errors.py
```diff
@@ -0,0 +1,40 @@
...
+    def testRecognizesUniqueViolation(self):
+        game = models.Game(game_id="th999", has_replays=True, num_difficulties=5)
+        game.save()
+
+        shot = models.Shot(game=game, shot_id="foo")
+        shot.save()
+
+        dupe_shot = models.Shot(game=game, shot_id="foo")
+
+        try:
```
jgdhs27, line 20: this looks like a job for [TestCase.assertRaises()](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises).

project/thscoreboard/shared_content/test_db_errors.py
```diff
@@ -0,0 +1,40 @@
...
+            )
+        else:
+            self.fail("Unexpected failure to raise an error")
+
+    def testDoesNotThinkOtherViolationIsUnique(self):
+        user = self.createUser("user")
+        replay = test_replays.CreateAsPublishedReplay("th6_extra", user=user)
+        replay.difficulty = -5
+
+        try:
```
jgdhs27, line 35: same as above

jgdhs27 (changes requested):
works nicely, just some small comments

project/thscoreboard/shared_content/test_db_errors.py
n-rook: I actually did this on purpose… for reasons that, a week later in hindsight, seem really silly. Done.

project/thscoreboard/replays/create_replay.py
n-rook: This is fair. I just added a classmethod for this. I didn't want to make a top-level constant called REPLAY_FILE_UNIQUE_HASH_CONSTRAINT or whatever. What I actually wanted to do was to either make a class-level constant or reference the constraint object directly, but I couldn't get either to work in a way I liked— the class-level constant couldn't be referenced in the declaration of the constraint, and similarly there wasn't a good way to get the constraint object from another file. So I just gave up and did this.

project/thscoreboard/replays/create_replay.py
n-rook: Eh… Since it's a private function I think it's fine. Were it a public function I would agree that either it should be a transaction or there should be a really big warning in the docstring not to call it outside of a transaction, and probably the former. But an extra transaction does have some overhead (even if it's pretty slight) so I think it's better to avoid it when possible.

n-rook: Looking at this function now I'm not sure why r.save() is even in here, so I moved it out. (But this conversation is still relevant, since the function still isn't safe outside of a transaction.)

jgdhs27: I like this approach

jgdhs27 (approved):

PR #469, opened 2023-09-26
Update ksy property name to match with generated files

hoangcaominh:
Probably not a very useful commit but it helps those who want to compare the `ksy` files with the generated files on thscoreboard ~~because the user generated files may not be the same as your generated files~~.

### DIFF
--- a/ref/threp-ksy/th11.ksy
+++ b/ref/threp-ksy/th11.ksy
@@ -37,7 +37,7 @@ types:
         size: 8
   stage:
     seq:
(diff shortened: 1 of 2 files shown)


jgdhs27 (approved):
tbh the fact that these dont match scares me. this is definitely a step in the right direction.

n-rook:
thanks btw, honestly the fact that the "generated files" don't exactly match what's generated by kaitai is one of the scarier parts of the codebase to me

PR #470, opened 2023-09-29
style button

jgdhs27:
Add `cursor: pointer` to clickable buttons. Add a new color when you hover over them.

### DIFF
--- a/project/thscoreboard/shared_content/static/style.sass
+++ b/project/thscoreboard/shared_content/static/style.sass
@@ -362,7 +362,7 @@ th, td
                 padding-left: 0
             &:last-child
                 padding-right: 0
-            
+
         .replay-heading-separator
             &::after
                 content: "•"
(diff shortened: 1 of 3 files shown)


project/thscoreboard/shared_content/static/style.sass
```diff
@@ -494,6 +494,10 @@ input
   font-size: 16px
   text-align: center
   margin-right: -5px
+  cursor: pointer
+
+.button:hover
+  background-color: var(--hover-button-color);
```
n-rook, line 500: You don't need semicolons in SASS.

n-rook (approved):
Thanks!

project/thscoreboard/shared_content/static/style.sass
jgdhs27: fixed

PR #471, opened 2023-10-06
Reorganize game list on left navigation bar.

n-rook:
It's hard to read a list of 10+ games without being organized, so this splits the list up a bit.

The actual categorization is fairly arbitrary, which is why the categories aren't labeled on the frontend. Historically, the English community has divided the Windows games into "Classic", "Modern1" and "Modern2" categories, where "Modern1" starts with Mountain of Faith and "Modern2" includes more recent engine upgrades. However, there isn't an obvious place to start "Modern2", since the engine upgrades were added over multiple games. As such, I chose to use the plot; MoF-TD have a clear thematic throughline, whereas DDC is totally unrelated to TD and serves as a new beginning.

I thought about styling the horizontal rule, but I wound up preferring the default grey to any color I came up with.

This does not affect the ranking page.

Screenshot 2023-10-06 at 1 47 58 PM

### DIFF
--- a/project/thscoreboard/thscoreboard/templates/base.html
+++ b/project/thscoreboard/thscoreboard/templates/base.html
@@ -61,10 +61,13 @@
         </div>
         <div class="content-and-sidebar">
             <nav class="sidebar">
-                {% for game_category, games in all_games_by_category.items %}
                 <h1 class="highlight">
-                    {{game_category}}
+                    {% translate "Scoreboards" %}
                 </h1>
+                {% for game_category, games in all_games_by_category.items %}
+                {% if not forloop.first %}
+                <hr>
+                {% endif %}
                 <ul class="game-list">
                     {% for game in games %}
                     <li>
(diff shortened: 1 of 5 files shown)


project/thscoreboard/thscoreboard/templates/base.html
```diff
@@ -61,10 +61,13 @@
...
         <div class="content-and-sidebar">
             <nav class="sidebar">
-                {% for game_category, games in all_games_by_category.items %}
                 <h1 class="highlight">
-                    {{game_category}}
+                    {% translate "Scoreboards" %}
                 </h1>
+                {% for game_category, games in all_games_by_category.items %}
+                {% if not forloop.first %}
+                <hr>
```
jgdhs27, line 69: could you maybe make the `<hr>` a bit thinner/lower contrast on dark mode? right now it stands out more than the text

jgdhs27 (approved):
Cool, I think this is a good idea and agree with how you split it.

PR #473, opened 2023-10-08
Replace emoji flags with SVGs.

n-rook:
Emoji flags aren't available everywhere; notably, Windows doesn't have them.


n-rook:
Screenshot 2023-10-08 at 5 18 32 PM

jgdhs27 (changes requested):
The US flag looks like this for me:

image

jgdhs27:
> The US flag looks like this for me:
> 
> image

(The Japanese flag looks correct)

n-rook:
Which browser do you use, so I can look into this?

The US flag is "really" much bigger than the Japanese one in the images, but it really shouldn't matter since I'm styling it with CSS (and it's SVG). So it must be a CSS browser issue for me to look into

jgdhs27:
> Which browser do you use, so I can look into this?
> 
> The US flag is "really" much bigger than the Japanese one in the images, but it really shouldn't matter since I'm styling it with CSS (and it's SVG). So it must be a CSS browser issue for me to look into

Chrome on desktop

n-rook:
Hmm... looks fine on my Windows machine too

32th-System:
It looks fine for me as well. It even shows up in the correct size in Internet Explorer
image

jgdhs27:
> Hmm, I'm on Chrome too (on a mac).
> 
> Could you try a cache-invalidating refresh? If that doesn't work, could you upload an image of the CSS when inspecting the img element in Chrome? That is, this:
> 
> Screenshot 2023-10-09 at 11 37 07 AM
> I'll try it on my Windows machine too when I get home.

Oh, it works as soon as i invalidate the cache. Sorry about that!

jgdhs27 (approved):

n-rook:
npnp, easy mistake to make. I've definitely been there

PR #475, opened 2023-10-08
Refactor setup_constant_helpers

jgdhs27:
Refactor this function, in order to allow us to make updates to constants as long as we don't change the primary key.

The motivation for this is that in the future, I want to make 'character' and 'subshot' be fields in each game, instead of things that we have to compute.

### DIFF
--- a/project/thscoreboard/replays/management/commands/setup_constant_tables.py
+++ b/project/thscoreboard/replays/management/commands/setup_constant_tables.py
@@ -316,36 +316,32 @@ def create_or_update_games(all_game_constants: List[GameConstants]):
 
 def create_or_update_shots(game=models.Game, shots=List[str]):
     shot_rows = models.Shot.objects.filter(game=game).all()
-    for shot_row in shot_rows:
-        if shot_row.shot_id not in shots:
+    if shot_rows:
+        shot_ids_in_db = {shot.shot_id for shot in shot_rows}
+        if shot_ids_in_db != set(shots):
             raise Exception(
-                f"Found shot {shot_row.shot_id} for game {game.id} in database, "
-                "but this shot is not in the list of shots for that game."
+                "The shots in the constant tables did not match the shots in the db"
             )
 
     for shot in shots:
         shot_row = models.Shot.objects.filter(shot_id=shot, game=game).first()
-        if shot_row:
-            pass
-        else:
+        if shot_row is None:
             shot_row = models.Shot(game=game, shot_id=shot)
             shot_row.save()
 
 
 def create_or_update_routes(game=models.Game, routes=List[str]):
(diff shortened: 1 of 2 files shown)


project/thscoreboard/replays/management/commands/setup_constant_tables.py
```diff
@@ -316,33 +242,107 @@ def _Create03():
...
+        create_or_update_routes(game_row, game_constants.routes)
+        game_row.save()
+
+
+def create_or_update_shots(game=models.Game, shots=List[str]):
+    shot_rows = models.Shot.objects.filter(game=game).all()
+    if shot_rows:
+        shot_ids_in_db = {shot.shot_id for shot in shot_rows}
+        if shot_ids_in_db != set(shots):
+            raise Exception(
```
n-rook, line 322: Rather than Exception, raise a more specific error here. I'd probably define a new exception subclass in setup_constant_tables, since none of the built-in ones really fit. This isn't so that you can catch the error (since this error should never really happen in the first place, and if it does happen there's nothing the app can do about it either) but instead is mostly so that the tests can be made more specific— assertRaises(Exception) could very easily pass by accident on all sorts of random errors, so it'd be very easy for somebody else like me to go in and break this by accident two years down the line, and not notice.

project/thscoreboard/replays/management/commands/setup_constant_tables.py
```diff
@@ -316,33 +242,107 @@ def _Create03():
...
+        else:
+            game_row = models.Game(
+                game_id=game_constants.id,
+                has_replays=game_constants.has_replays,
+                num_difficulties=game_constants.num_difficulties,
+            )
+
+        create_or_update_shots(game_row, game_constants.shots)
+        create_or_update_routes(game_row, game_constants.routes)
+        game_row.save()
```
n-rook, line 314: I'm actually surprised this works— I would have thought the shots and routes would have to be saved after the Game row, since otherwise the foreign key wouldn't resolve properly. But I guess it does work; I was wrong!

n-rook (approved):
Nice! This'll be great to have; there's a lot of janky stuff we do right now because we don't have this capability.

project/thscoreboard/replays/management/commands/setup_constant_tables.py
jgdhs27: interesting, i didnt even really think about that

project/thscoreboard/replays/management/commands/setup_constant_tables.py
jgdhs27: good idea, done

n-rook (approved):

PR #477, opened 2023-10-09
Annotate base.html strings for translation.

n-rook:
I found out recently that apparently the word wrapping in these files has no significance whatsoever. I'm going to have to figure out how to get it to stay consistent later so I stop getting these giant diffs for no reason.


32th-System (approved):

n-rook:
Thanks for the reviews!

PR #479, opened 2023-10-10
Update repo to match new, Debian Bookworm state of the server.

n-rook:
There are a few different types of changes in this PR:

- The new version of pip supplied strongly prefers venv. As such, to get it working, I started using venv. As such, I document it here and update the uwsgi.ini files accordingly.
- uwsgi-staging was crafted outside of git, so I bring it in here.
- uwsgi is now installed through apt (though I'm not sure how it was installed before). As such, we now run a binary called uwsgi_python311 which is a version of uwsgi built to support Python 3.11. (I'm not sure exactly what the deal is with uwsgi's plugin architecture, but this works for now.)
- The new Debian supplies Python 3.11, so I started using it.
- A few Python libraries got accidentally left out of requirements.txt. requests is used by the Discord plugin that announces new versions. django-compressor is used by the `compilescss` admin command.

Normally I would put these in different PRs, but since this all constitutes one big update to bring over the changes I've already done on the prod server, I figured it was easiest to see it all as one big piece.

### DIFF
--- a/.github/workflows/run_python_tests.yml
+++ b/.github/workflows/run_python_tests.yml
@@ -31,7 +31,7 @@ jobs:
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v3
       with:
-        python-version: "3.10"
+        python-version: "3.11"
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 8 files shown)


32th-System:
uwsgi was previously installed using pip, to ensure that we'd have the latest version (with the latest security fixes) straight from the developers. The Debian version should be fine for security though

32th-System (approved):

n-rook:
oh yeah that makes sense. I just kinda acted quickly because I accidentally broke the site :p

PR #480, opened 2023-10-12
Create a test for the "create replay" view.

n-rook:
This test is largely a proof-of-concept: I want to add a new view but was worried enough about it that I wanted a test for it. So, I figured I would add a test for an existing view first, to see how it worked.

In the past, we had a test like this and it was flaky enough that we got rid of it. However, this new test differs from the old one in that it uses RequestFactory:

https://docs.djangoproject.com/en/4.2/topics/testing/advanced/

My understanding is that this means that no actual HTTP requests are being issued; rather, we are constructing a fake request object directly in the test. This should avoid the source of flakiness.

I'm a bit ambivalent about this test; it's unfortunate that I have to directly pass in a dictionary that is supposed to correspond to the form on the page. I do think this test has value, mainly to warn us if we accidentally totally break this page, but I'm not about to start mandating similar tests for every new view. In terms of the test hierarchy, this is really just a giant unit test, where the unit being tested is the view.

### DIFF
--- a/project/thscoreboard/replays/urls/replay_urls.py
+++ b/project/thscoreboard/replays/urls/replay_urls.py
@@ -12,7 +12,11 @@ urlpatterns = [
     path("", index.index, name="index"),
     path("index/json", index.index_json),
     path("upload", create_replay.upload_file, name="upload_file"),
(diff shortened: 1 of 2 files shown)


jgdhs27 (approved):
nice

PR #488, opened 2023-12-02
Lower batch size for reanalyzing replays.

n-rook:
5000 took around 90s, so this should take around 9s.

I also added an env variable to customize this behavior, so if we need to change it later we don't need to touch the source code.


n-rook:
oops I'm bad at math lol

PR #490, opened 2023-12-24
Add a management command to recompile the Kaitai structs.

n-rook:


project/thscoreboard/replays/management/commands/kaitai_struct_compile.py
```diff
@@ -0,0 +1,31 @@
+"""A command to regenerate the .ksy files."""
+
+import logging
+import pathlib
+import subprocess
+
+from django.core.management import base
+
+
+_PROJECT_ROOT = pathlib.Path(__file__).parent / "../../.."
```
jgdhs27, line 10: i dont like how `..` is mixed with `.parent` here, i would prefer just to use `.parent`

project/thscoreboard/replays/management/commands/kaitai_struct_compile.py
```diff
@@ -0,0 +1,31 @@
...
+        ksc_args = [
+            "kaitai-struct-compiler",
+            "--target",
+            "python",
+            "--outdir",
+            _PROJECT_ROOT / "replays/kaitai_parsers",
+        ] + _ksy_files()
+
+        logging.info("Running: %s", ksc_args)
+        subprocess.run(ksc_args)
```
jgdhs27, line 31: didn't work for me, i had to use 
`        subprocess.run(ksc_args, shell=True)`
(see also https://stackoverflow.com/questions/73193119/python-filenotfounderror-winerror-2-the-system-cannot-find-the-file-specifie)

im on windows btw

jgdhs27 (changes requested):

n-rook:
I don't like shell=True— it makes me nervous, weird stuff can happen if you let the shell handle things— so I'll close this for now and reopen when I get home and can figure something out with my Windows machine

Issue #438, opened 2023-07-30
Update site's Python to 3.11


n-rook:
We recently broke staging by accident by using a new Python feature.

We should update the site to 3.11. What would be ideal is if we figured out a way to update only staging first and then prod. But it wouldn't be the worst to just test the site locally and then update both staging and prod on the server.

32th-System:
The server now uses Python 3.11

PR #491, opened 2023-12-24
Delete unused runtime.txt and use pyenv file instead.

n-rook:
.python-version is used by pyenv to contain the Python version we should use. Note that neither of these actually do anything on the real web server.


jgdhs27 (changes requested):
I think we should be using python 3.11, see https://github.com/n-rook/thscoreboard/issues/438

n-rook:
Oops, you're totally right. Fixed.

Issue #61, opened 2022-09-19
Don't use edited generated Kaitai Struct files for early Touhou games to deal with stage offsets


raviddog:
#54 - see comments

PR #492, opened 2023-12-25
Update the Kaitai Struct definitions to no longer require manual edits to generated code.

n-rook:
Previously, we edited the generated code in order to work around an issue with our definitions or with the compiler (it's unclear which). Now, the definitions have been restructured so that this is no longer necessary.

#61

As always, no need to really review the generated code, just the definitions (and replay_parsing.py which is not generated code).


jgdhs27 (approved):
It's very nice to have this fixed!

n-rook:
I reverted #490 on this branch and changed the base branch to main so that I can merge it now— I won't be fixing #490 for at least a few days, so there's no need for this PR to wait on it.

PR #499, opened 2024-01-28
Redirect to the page you came from after logging in

### LINKED ISSUE #489: After logging in, it'd be nice to take the user to the page they were on
Currently, if the user is on some page, and then they click "Log in" and log in, we take them to the home page; we don't remember the page they were on. This is fine, but it'd be nice to take them to the page they were on before they logged in; many sites do that.

32th-System:
turns out this is already a feature of the login page, we just needed to make use of it

Also made sure to not redirect to /users/logout

Fixes #489


n-rook (approved):
Nice!

Looks like something is busted w/ the linter, I'll look into that

PR #501, opened 2024-01-28
Update requirements

32th-System:
fixes dependabot complaints


requirements.txt
```diff
@@ -1,12 +1,12 @@
-bcrypt==4.0.1
-django==4.2.4
+bcrypt==4.1.2
+django==5.0.1
```
n-rook, line 2: I'm scared of updating all this stuff in one PR, it seems like it could break something. Could we split it into at least 3? (requirements-dev.txt, requirements.txt [everything except django], and django)

n-rook:
Closing obsolete PRs

Issue #507, opened 2024-08-10
Inconsistent ranking between Recent Records and the game leaderboard


KirbyComment:
The recent records page can display an incorrectly labeled medal (placement) from what the real placement is as seen from the game leaderboard.

![image](https://github.com/user-attachments/assets/90f5a67d-dc50-471b-8fc4-b3177a025320)

![image](https://github.com/user-attachments/assets/db07d2ae-fe65-4f30-88ef-118b34e7c253)

n-rook:
Replay is:
https://www.silentselene.net/replays/th128/77823

Notably there's a stronger TAS:
https://www.silentselene.net/replays/th128/77752

As the front page displays TASes, I think there's something going wrong with the database query that is causing it to include TASes in the ranking too.

PR #508, opened 2024-08-19
Creates and joins against a view used to get the rank of a replay.

n-rook:
See #507. This new function is not yet used, but I figure we can replace the existing annotate_with_rank() with it. The new approach allows for us to more easily get ranks in situations where the required data isn't returned in the query, which is nice. It's a bit slower, so we may need to convert this to a "materialized view", but I think it's still acceptable right now, especially with the new index.


jgdhs27 (approved):
This is so much better! I didn't know you could just write views like this.

Btw, how did you define the view? Did you just write the SQL to create it in `0041_replay_rank_view.py`?

n-rook:
Yeah, I wrote the 0041 migration file myself. Django doesn't really support views in any meaningful sense, but a view can be queried the same way as a table, so if we make a view ourselves (and tell Django not to try to create it as a table; that's what the "managed = False" is for) we can access it through Django similarly to how we would access a table.

It took a bunch of Googling to figure out how people deal with this situation!

PR #509, opened 2024-08-24
Use new rank view everywhere the old annotate_with_rank call was used.

### LINKED ISSUE #507: Inconsistent ranking between Recent Records and the game leaderboard

n-rook:
This should fix #507, as well as simplifying some code to no longer need rank_replays.py.


jgdhs27 (approved):
Awesome! What's the performance implication of this? I imagine it's a lot better now that you don't have to calculate it from scratch every time (idk how view tables work, are they stored in the DB?). The medal calculation was actually really slowing down how long it took to show replays on the home page and player pages.

n-rook:
Unfortunately it's pretty much a wash for now. A view is basically just a query that the database pretends is a table. It didn't feel worse to me locally. I planned to click around on staging to see how it felt.

In the medium term what we should probably do is use a "materialized view" here:
https://www.postgresql.org/docs/current/rules-materializedviews.html

These are views that postgres saves to a table. The big caveat is that the table can then go stale and you have to go tell postgres whenever it should be updated. But this is a classic materialized view problem, since it's a small table with a big input and not that much information.

It'll be much easier to switch to a materialized view now that we have a regular view, so I figure this is still a step in the right direction. And it does fix the front page bug.

PR #523, opened 2025-05-10
add Japanese Translate

wefma:
To make the site more accessible and friendly to Japanese users, we will translate almost all of the items into Japanese.


wefma:
By removing "様" only from the site administrator and placing the other developers under the "Contributors" section, we can treat the contributors with '様' if we choose to.

wefma:
For the text visible to general users, I focused on conveying the meaning in a way that is easy to understand, taking into account the actual behavior. However, for legal documents such as the terms of service, I made an effort to use expressions as close as possible to the original meaning.

wefma:
I don't have a runtime environment to test it yet, but I'll open a pull request for now just in case.
if you ok, please merge this

project/thscoreboard/replays/templates/replays/index.html
```diff
@@ -4,10 +4,15 @@
 {% block content %}
 
 <h1 class="highlight">{% translate "Welcome to Silent Selene!" %}</h1>
+{% blocktranslate %}
```
n-rook, line 7: `{% blocktranslate %}` cannot be used outside of a block with `{% url ... %}` in it. As such, please revert this change from this PR, and submit it separately later.

I wanted to be helpful and do this myself now that you pointed it out. However, when I did so and then ran "makemessages" it caused a lot of small differences in the .po files. If I submitted that, I think it would cause many merge conflicts on your PR. So, I think the best way to handle this is:

1. You remove this annotation fix in the index page from this PR.
2. I approve this PR and submit it.
3. I fix this annotation problem myself.
4. Once I do, you can submit the translation for this block of text.

Sorry for the inconvenience.

project/thscoreboard/locale/ja/LC_MESSAGES/django.po
```diff
@@ -1609,6 +1800,96 @@ msgid ""
...
+"    <p>\n"
+"        サイレントセレナのリプレイは『通常プレイ』『ツール使用』『その他』の3種類に分類されます。多くのリプレイは<b>通常プレイ</b>に分類されます。通常プレイはプレイヤーに作用する特殊なツールを用いてプレイしたものではないことを意味します。(ただしプレイ内容に変更がないツールは通常プレイに分類されます。)<b>ツール使用</b>のリプレイは外部ツールを使用してプレイされたリプレイのことを指します。任意のツールが該当し、ステートセーブのように明らかに『不正』とみなされる行為が行われているリプレイであっても許されます。しかし、そのリプレイは実際のゲームで動作するリプレイでなければなりません。そして<b>その他</b>は上記2つに該当しないもの全てのリプレイを指します。これはゲームの内容を改変したリプレイや、その他著しく通常の挙動から逸脱した変更を加えたリプレイが該当します。\n"
+"    </p>\n"
+"    <h2>要約</h2>\n"
+"    <p>\n"
+"        通常プレイではVpatchと翻訳パッチが許可されています。ポーズバッファリング(一時停止を濫用して疑似的にゲーム速度を遅くすること)は許可されていません。\n"
+"    </p>\n"
+"    <h2>自分のプレイが通常プレイに当てはまるか否か</h2>\n"
+"    <h3>一般規則</h3>\n"
+"    <h4>高難度プレイのリプレイ\</h4>\n"
```
n-rook, line 1816: The backslash before `</h4>` causes the translation library to report an error, so please remove it.

n-rook (changes requested):
Thank you so much for doing all of this! I really appreciate it. I have felt bad that the Japanese translation has been incomplete for so long, so I am very happy to see it improve.

project/thscoreboard/replays/templates/replays/index.html
wefma: Is this ok?
[d088e11](https://github.com/n-rook/thscoreboard/pull/523/commits/d088e11b67d805614063b180c0c3d5bbbce14840)

Sorry for the trouble on my end too

project/thscoreboard/locale/ja/LC_MESSAGES/django.po
wefma: sorry for type miss!
[ed03bee](https://github.com/n-rook/thscoreboard/pull/523/commits/ed03bee37e79cfcaa7bd16177b281c35236f8225) modify this wrong

project/thscoreboard/replays/templates/replays/index.html
n-rook: Thank you. I had to remove it from the .po file too to get it to work, but I pushed that changes so it should be good now.

project/thscoreboard/locale/ja/LC_MESSAGES/django.po
n-rook: Thanks

n-rook (approved):

PR #541, opened 2025-08-27
ja-translate day1 support for th20

wefma:
Sorry for cannot check to run silent selene in Japanese translate in my environment...
Could you check what it run ...?


n-rook:
Screenshot 2025-08-27 at 3 38 10 PM

Seems like it looks good. Thank you

Issue #531, opened 2025-06-07
GFW stage practice appears as full game


n-rook:
https://www.silentselene.net/replays/th128/80121

This looks like stage practice, but it's appearing as full game for some reason.

hoangcaominh:
It is stage practice, except it's a thprac thing in GFW.
The parser for GFW always returns the replay type as full game because nobody thought someone would upload a GFW stage practice replay using thprac 🤷

PR #546, opened 2025-11-29
add Uwabami Breakers

wefma:
rejecting this pull request is ok if admin think that Uwabami Breakers is not in scope in silentselene


wefma:
https://github.com/n-rook/thscoreboard/issues/531
this related this issue because th128 and alco don't have practice mode.

n-rook (approved):
This looks good to me. Thank you very much.

I think I will leave it on staging for a little while and ask people to try it out before deploying it to production. But I see no obvious problems.

PR #556, opened 2026-04-26
add th095

wefma:
~because avoiding conflict, the base branch is wefma/alco~
I rebased to the main because wefma/alco is merged into main


project/thscoreboard/replays/game_ids.py
```diff
@@ -662,7 +674,28 @@ def GetRouteName(game_id: str, route_id: str):
...
+
+
+def GetSceneGameSceneName(game_id: str, scene_game_scene: int | None) -> str:
+    if game_id == GameIDs.TH095:
+        if scene_game_scene is not None:
+            return str(scene_game_scene)
+    return "Bug scene"
+
+
+def GetDifficultyName(
```
n-rook, line 699: Is there any way we could add a new function for this instead of difficulty? I get why you would want to display this in the difficulty column on in tables, but I think it is too confusing to have "GetDifficultyName" return which scene is being played. It is too different from what difficulty is in other games.

n-rook (commented):
Thank you for this contribution. Sorry that I have not had time to give it a full review yet. Supporting scene games is very complicated, which is why I got a third of the way through supporting them (with the .ksy) and then gave up and put it off. I did have one initial comment to leave.

project/thscoreboard/replays/game_ids.py
wefma: Sorry, I’ll fix it.
I’ll add a separate function for the scene name and use it for the Difficulty column in the table.

n-rook (approved):
Thank you a lot. I tried this out locally and it seemed to work well. The scene game players will be happy too. I will merge this to staging now, but will wait for the weekend to push it to production

wefma:
thanks!

Issue #561, opened 2026-04-29
If browser prefers Japanese to English, can't switch to English language (due to invalid `en_US` language code)


wefma:
Django expects `en-us` in `LANGUAGE_CODE`, `LANGUAGES`, and the language cookie.
https://docs.djangoproject.com/en/6.0/ref/settings/

wefma:
[diff.patch](https://github.com/user-attachments/files/27209147/diff.patch)
maybe apply this diff.patch?
Its validity is unverified.

wefma:
you can't switch language code.
this bug caused by this, maybe

n-rook:
What exactly is the bug? I can switch language code on the main website or on staging. Does it not work on some machines? Or is it that it is broken in some way that is harder for me to notice?

PR #562, opened 2026-04-30
fix i18n

wefma:
resolve this issue: https://github.com/n-rook/thscoreboard/issues/561


wefma:
I changed the language selection from buttons to a select dropdown.

wefma:
We should add i18n support to other pages like login.html.
However, applying it to all pages would require significant changes.
So I only fixed the language selection functionality and filled in missing strings in the existing locale files.

n-rook:
Sorry, I am a little confused.

- What is the bug that is being fixed by this change? When I click the flag in the top right, the site does switch between English and Japanese, as is expected.
- I think the flag button is fine as long as we only support English and Japanese, and that a drop down would be harder to use. If we plan to support additional languages, we would want a more complicated way to switch languages, but I do not think we need that yet.

wefma:
I’m using Silent Selene in Google Chrome with my browser language set to Japanese. In that environment, switching the site language from Japanese to English does not work correctly.

My understanding is that the form currently submits `en_US`, which is then stored in the language cookie. However, Django’s configured language code is `en-us`, not `en_US`, so Django does not recognize the cookie value as a supported language. It then falls back to the browser’s `Accept-Language` header. Since my Chrome is configured to prefer Japanese, Django selects `ja` again, so the page stays in Japanese.

This may be why it appears to work in some environments: if the browser already prefers English, the fallback still resolves to English. But in browsers/profiles where Japanese is preferred, the language switch fails.

That said, I understand your concern about the dropdown. I agree that while the site only supports English and Japanese, the current flag button UI is probably simple enough and easier to use.

If the dropdown change is not desirable, I’m happy to revert that part of the PR and keep the existing flag button behavior. The important part of the fix is changing the submitted English language code from `en_US` to `en-us`.

https://www.youtube.com/watch?v=1VlqPUVBELQ

n-rook:
Okay, I have confirmed that if your browser requests Japanese in favor of English, then the flag button does not work. I believe you that the cause is the bug of using the wrong language code. Thank you very much for reporting this. I feel sorry for the inconvenience to Japanese users.

I want to keep the flag for now, so if you don't mind, it would be great if you would revert the change to the button and keep the flag, just switching the language code. (That is, what you already suggested.)

wefma:
> `[bff43ad](/n-rook/thscoreboard/pull/562/commits/bff43ad469069198a12abec0d1a2d0511a322637) `
> [ ](/wefma)

ok, I reverted in [7eab0b1](https://github.com/n-rook/thscoreboard/pull/562/commits/7eab0b175ba3b9addae719c4b6cf769d4bfed1e1) and, 
fix it in [bff43ad](https://github.com/n-rook/thscoreboard/pull/562/commits/bff43ad469069198a12abec0d1a2d0511a322637).


And fixed locale file in [9b44b9d](https://github.com/n-rook/thscoreboard/pull/562/commits/9b44b9d0d6b3e292d933a87f4eea54a58952da42)
Maybe there are no untranslated msgids remaining.

```bash
wefma@wefma:~/ghq/github.com/n-rook/thscoreboard/project/thscoreboard$ msgfmt --statistics locale/ja/LC_MESSAGES/django.po
431 translated messages.
wefma@wefma:~/ghq/github.com/n-rook/thscoreboard/project/thscoreboard$ msgfmt --statistics locale/en_US/LC_MESSAGES/django.po
431 translated messages.
wefma@wefma:~/ghq/github.com/n-rook/thscoreboard/project/thscoreboard$
```

n-rook (approved):
Thank you!

PR #572, opened 2026-06-08
Format it correctly as an application using Python 3.11

wefma:
The purpose of this change is to standardize the codebase on Python 3.11, since Python 3.10 is currently mixed in, and to update code that uses patterns deprecated or discouraged in Python 3.11.


wefma:
I opened this pull request.

n-rook (approved):
Makes sense to me. Note that it is Python 3.11 that is deployed on the production server.

PR #574, opened 2026-06-09
using devcontainer

wefma:
I hate recreate postgres tables.
So I want to create easily to destroy and recreate database and other environment.


n-rook:
FWIW I personally don't mind just running everything myself, but I am happy to merge this in once you have it working well and think it is ready to merge

PR #576, opened 2026-06-15
Use npm ci with a committed lockfile for JS tests

wefma:
The GitHub Actions workflow was running both `npm install` and `npm ci`, causing dependencies to be installed twice.
This change commits `package-lock.json` and updates the workflow to run only `npm ci`, so JavaScript test dependencies are installed reproducibly from the lockfile.


n-rook (approved):
Looks good. Thanks

PR #578, opened 2026-06-24
Add package hashes to requirements texts

wefma:
Pinned versions alone do not verify the integrity of downloaded packages. Hash verification ensures that pip installs only the expected package artifacts and provides additional protection against supply-chain attacks.


n-rook:
Hmm. Does this approach really make us more secure?

I understand the argument that it protects against supply-chain attacks. But realistically, when we generate these hashes to check into the codebase, we are not manually inspecting each line of each dependency to make sure it makes sense and does not contain anything dangerous. So we are still trusting pip anyway, right?

wefma:
I agree with that. We are still trusting pip when the hashes are initially generated.

What I want to guarantee is different.
If we install a non-malicious version of a package during development, then production, CI, and other developers can install those exact same package, rather than whatever happens to be available later.
In other words, the hashes ensure that everyone installs the exact packages that were originally reviewed and committed.

Also, most large-scale supply-chain attacks tend to be discovered and reported within a few days.
Though that is certainly not a guarantee of safety, it does mean that the risk is significantly lower if a newly introduced dependency has been publicly available for some time without any reports of compromise.
Using hashes lets us keep installing that known packages instead of silently picking up a different one later.

PR #579, opened 2026-06-26
add workflow permissions

wefma:


n-rook:
Doesn't it have these permissions by default anyway? (Otherwise it wouldn't work at all, right?)

wefma:
I may have chosen the wrong title for this pull request...

What I intended wasn't to add permissions, but to explicitly restrict the workflow to the minimum permissions it needs.

This follows the principle of least privilege and helps prevent any code executed during the workflow, such as a compromised dependency or GitHub Action, from using the workflow's credentials to write to the repository.

n-rook:
Ah, that makes sense. Actually, I had a difficult time finding the docs for this feature, but I'm aware that a lot of repos have been attacked through GitHub Actions, so I appreciate you looking into this.

n-rook (approved):

PR #584, opened 2026-08-10
Parse V14 replays from the PoDD replay patch

christianazinn:
(Edited) This PR adds parsing support for the V14 replay format from the PoDD replay patch (https://azinn.dev/th03/th03-data-manager-latest.zip), which is currently in beta. This supersedes the V13 format previously proposed here. V11, V12, and V13 replays can be converted with Data Manager 0.5.6.

As requested, this PR only adds the Kaitai structs and `replay_parsing.py` support. It handles Story and VS/netplay replays, though not practice (although I don't know why you'd want to upload a practice replay?) The application-side changes are in #585.


n-rook:
This is still in progress, but I'll add a few notes so you know what I'll be saying when it's ready to review:

- From a code perspective, I recommend splitting this into at least two PRs; one to add parsing (that is, Kaitai structs and replay_parsing.py) and then one to handle the app side of things
- From a product perspective, it's absolutely required that it still be possible to create PoDD scores without a replay (for obvious reasons). So the handling will have to be a little different from how the windows games work

Issue #586, opened 2026-09-02
th20 (Fossilized Wonders) replay format has been reverse-engineered


hakatashi:
Hi! I recently launched [TouhouSattori](https://sattori.hakatashi.com/en), a project that automatically records Touhou replay files. Using th20 replay data and recording metadata collected by that project, I've reverse-engineered the th20 replay file format almost completely (with substantial help from AI). I've written up the details here:

https://github.com/hakatashi/sattori/blob/main/docs/research/th20-replay-format.en.md

As far as I can tell, no prior public analysis of th20's replay body format — in particular the per-stage split structure — exists anywhere. I also checked this project's parser and confirmed that replays/kaitai_parsers/th20.py currently defines only a Header class, with no Stage class. This matches what I see on every th20 replay page on Silent Selene: "Stage split information is unavailable for this replay."

I think this could be useful for thscoreboard. Please consider implementing a proper th20 replay parser with stage-split support. If it would help, I'd also be happy to implement the per-stage parsing myself and send a pull request.

Disclaimer: I currently use Claude Code for nearly all of my coding work. That said, I always have a human review all code before sending patches to external projects, and I understand that I'm fully responsible for any code submitted under my account.

hakatashi:
As a side note, I also used AI to analyze the replay files for new versions of th06 (both the classic and new classic versions) in a similar way:
https://github.com/hakatashi/sattori/blob/main/docs/research/th06-classic-replay-format.en.md

raviddog:
> As a side note, I also used AI to analyze the replay files for new versions of th06 (both the classic and new classic versions) in a similar way: https://github.com/hakatashi/sattori/blob/main/docs/research/th06-classic-replay-format.en.md

Thanks, this is helpful

PR #588, opened 2026-09-04
Add TH06 New Classic support

raviddog:
opening the PR so people can test and check that I remember how to do this properly


raviddog:
Added:
- TH06NC replay decoding + version check
- Sidebar link to game
- Replay stage "misses" field
- NC_CHALLENGE as new Replay Type for the added Challenge mode
- game fields check for the NC_CHALLENGE replay type
- translations for new game strings
- migrations for new db fields/values

raviddog:
I wrote some tests, but I haven't committed them because I don't want to publish the replay files until the game is released. They all pass though.

raviddog:
- Updated TH06NC parser to support launch version
  - added stage practice replay support
  - #586 , thanks @hakatashi
- Add parser support for TH06c
- Add version field to replays to distinguish between 1.02 and 1.03 TH06 replays

project/thscoreboard/replays/test_replay_parsing.py
```diff
@@ -640,3 +640,18 @@ def testAllClear(self) -> None:
         s1end = r.stages[0]
         self.assertEqual(s1end.stage, 1)
         self.assertEqual(s1end.score, 1269907)
+
+
+class Th06CReplayTestCase(unittest.TestCase):
```
n-rook, line 686: I think we probably need a lot more tests here just to make sure it works in different cases. I would at least think we need:

1. A few more asserts on the classic replay, just to check that it is parsed right
2. A NC full game replay
3. A NC challenge mode replay

Be sure to check all of the parsed fields too. I checked my own NC replay and the stage split scores (and power) didn't look quite right.

project/thscoreboard/thscoreboard/templates/base.html
```diff
@@ -71,6 +71,7 @@ <h1 class="highlight">
                 <ul class="game-list">
                     {% for game in games %}
                     <li>
+                        {{game.GetSidebarPrefix}}
```
n-rook: IMO let's hold off on the sidebar symbol for now. We can add it in later. I have a few minor concerns with the implementation and I think it makes sense for us to get this PR in before you have to respond to me going "but wait, the CSS isn't quite right" :)

FWIW my concerns are:
1. I think there should be spacing between the indent symbol and the icon
2. I think the indent symbol shouldn't be selectable. For example right now, an I-bar cursor will appear if I hover over it, which just doesn't look quite right

n-rook (commented):
Looks mostly good. Thanks so much for doing this

project/thscoreboard/thscoreboard/templates/base.html
raviddog: I removed it for now

project/thscoreboard/replays/test_replay_parsing.py
raviddog: Yeah I made a change to the parser and forgot to update the .ksy file accordingly. That's been fixed and I added a bunch of tests to cover some cases I thought of.

raviddog:
- rebased? I think? sorry for commit history?
- update th06nc ksy to match a parsing change I made earlier
- remove sidebar indent
- added more tests

raviddog:
- added data structure for TH06NC spell card names. I haven't added the actual names yet, just the IDs so people can easily add in the correct information later

n-rook (approved):
Thanks!

PR #599, opened 2026-09-15
modify EoSD NC spell names

nit-alcostg:
modify EoSD NC spell names


nit-alcostg:
I deleted the last three spells because we cannot play these spells in spell card practice.

n-rook (approved):

n-rook:
Thank you!

PR #600, opened 2026-09-16
add gitattributes

wefma:
Some existing text files use CRLF line endings.
This change normalizes them to LF and enforces LF line endings for future commits.


wefma:
spell_names.py uses CRLF too.
However, if this python file add now, this pull request will conflict https://github.com/n-rook/thscoreboard/pull/599
so I will rebase this pull request after https://github.com/n-rook/thscoreboard/pull/599 is merged.

